Issue #5, opened 2018-01-12
Don't overwrite appsscript.json on create


sqrrrl:
Noticed that create does a fetch of the newly created project and overwrites any local files. In some cases that's undesirable (e.g. clone a project from git, create a new project before pushing code.) Would be helpful to skip the fetch if current directory looks like a valid project (e.g. appsscript.json present)

grant:
Will close after published to npm.

PR #6, opened 2018-01-12
Skip fetching project on create if manifest already exists

sqrrrl:
Was trying to add clasp instructions for the gmail add-on samples and noticed that creating a project overwrites the current manifest. For cases where somebody is cloning via git instead of from an apps script project, we shouldn't overwrite things.

Also fixed the license name in the module metadata to avoid warnings from NPM.


sqrrrl:
Corresponds to issue #5

grant:
LGTM in general, I'll test it once back to work.
Next time, we should make the small license change in a different PR.

PR #19, opened 2018-01-20
Fix #13, streamline spaces in object destructuring

### LINKED ISSUE #13: Remove fs write warning when fetching
There's a warning showing when fetching a project:

**Warning**: `(node:46497) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.`
**Node version**: v9.0.0
**Fix**: In `fetchProject`, use a proper callback:
```
            fs.writeFile(`./${filePath}`, file.source, (err) => {
              if (err) console.error(err);
            });
```

Nirazul:
This PR fixes #13 and removes the logging of  `null` in the latest version of code.
It also contains a bit of code style streamlining. As there is no eslint to prove what is desired, I can revert this, if you would like to keep the current version.


Nirazul:
I signed it!

PR #26, opened 2018-01-24
add parent id parameter to create command and enable user to bind script with any supported  document

bladechild:
add parent id parameter to "create" command and enable user to bind script with any supported  document

### DIFF
--- a/index.js
+++ b/index.js
@@ -104,10 +104,10 @@ const DOTFILE = {
 // @see https://developers.google.com/oauthplayground/
 const REDIRECT_PORT = 2020;
 const oauth2Client = new OAuth2(
-    '1072944905499-vm2v2i5dvn0a0d2o4ca36i1vge8cvbn0.apps.googleusercontent.com', // CLIENT_ID
-    'v6V3fKV_zWU7iw1DrpO1rknX', // CLIENT_SECRET
-    'http://localhost:' + REDIRECT_PORT
-    // 'urn:ietf:wg:oauth:2.0:oob' // REDIRECT_URI (@see OAuth2InstalledApp)
+  '1072944905499-vm2v2i5dvn0a0d2o4ca36i1vge8cvbn0.apps.googleusercontent.com', // CLIENT_ID
+  'v6V3fKV_zWU7iw1DrpO1rknX', // CLIENT_SECRET
+  'http://localhost:' + REDIRECT_PORT
+  // 'urn:ietf:wg:oauth:2.0:oob' // REDIRECT_URI (@see OAuth2InstalledApp)
 );
 const script = google.script({
   version: 'v1',
@@ -323,7 +323,7 @@ program
 
       // Create a local HTTP server that reads the OAuth token
       var app = connect();
-      app.use(function(req, res) {
+      app.use(function (req, res) {
         var url_parts = url.parse(req.url, true);
         var code = url_parts.query.code;
         if (url_parts.query.code) {
@@ -339,7 +339,7 @@ program
         res.end(LOG.AUTH_PAGE_SUCCESSFUL);
       });
       http.createServer(app)
-          .listen(REDIRECT_PORT);
+        .listen(REDIRECT_PORT);
     });
   });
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -354,20 +354,21 @@ program
...
-  .action((title = LOG.UNTITLED_SCRIPT_TITLE) => {
+  .action((title = LOG.UNTITLED_SCRIPT_TITLE,parentId) => {
     if (fs.existsSync(DOT.PROJECT.PATH)) {
       logError(null, ERROR.FOLDER_EXISTS);
     } else {
       getAPICredentials(() => {
         spinner.setSpinnerTitle(LOG.CREATE_PROJECT_START(title));
         spinner.start();
-        script.projects.create({ title }, {}, (error, res) => {
+        script.projects.create({ title,parentId }, {}, (error, res) => {
```
grant: space after ,

index.js
```diff
@@ -354,20 +354,21 @@ program
...
+ * Creates a new script project. The project title is optional. The project parent Id is optional.
+ * ParentId: The Drive ID of a parent file that the created script project is bound to. This is usually the ID of a Google Doc, Google Sheet, Google Form, or Google Slides file. If not set, a standalone script project is created.
  * @example `create "My Script"`
  */
 program
-  .command('create [scriptTitle]')
+  .command('create [scriptTitle] [scriptParentId]')
   .description('Create a script')
-  .action((title = LOG.UNTITLED_SCRIPT_TITLE) => {
+  .action((title = LOG.UNTITLED_SCRIPT_TITLE,parentId) => {
```
grant: space after ,

index.js
```diff
@@ -354,20 +354,21 @@ program
   });
 
 /**
- * Creates a new script project. The project title is optional.
+ * Creates a new script project. The project title is optional. The project parent Id is optional.
+ * ParentId: The Drive ID of a parent file that the created script project is bound to. This is usually the ID of a Google Doc, Google Sheet, Google Form, or Google Slides file. If not set, a standalone script project is created.
```
grant: Please make line length <= 80.

index.js
```diff
@@ -354,20 +354,21 @@ program
   });
 
 /**
- * Creates a new script project. The project title is optional.
+ * Creates a new script project. The project title is optional. The project parent Id is optional.
```
grant: Now that we have multiple parameters, can you describe them roughly like this?

```js
/**
 * Creates a new script project.
 * @param {string} [scriptTitle] An optional project title.
 * @param {string} [scriptParentId] An optional project parent Id. The Drive ID of a parent file
 *   that the created script project is bound to. This is usually the ID of a
 *   Google Doc, Google Sheet, Google Form, or Google Slides file. If not set, a
 *   standalone script project is created.
 * @see https://developers.google.com/apps-script/api/reference/rest/v1/projects/create#body.request_body.FIELDS.parent_id
 */
```

index.js
```diff
@@ -354,20 +354,21 @@ program
...
+ * ParentId: The Drive ID of a parent file that the created script project is bound to. This is usually the ID of a Google Doc, Google Sheet, Google Form, or Google Slides file. If not set, a standalone script project is created.
  * @example `create "My Script"`
```
grant: Can you please add another line to include an example with a scriptParentId?

index.js
bladechild: Sure, I will do it. Thanks

index.js
bladechild: Thank you, I will do it

index.js
bladechild: Yes, I will add an example soon

index.js
bladechild: Thank you

index.js
bladechild: Thank you

index.js
```diff
@@ -323,7 +323,7 @@ program
 
       // Create a local HTTP server that reads the OAuth token
       var app = connect();
-      app.use(function(req, res) {
+      app.use(function (req, res) {
```
grant: Don't change this next time. (We should use an arrow function if we change).

bladechild: Thanks. The space after function is actually added by Eslint. I will pay more attention next time.

PR #27, opened 2018-01-24
Save files as .js. Treat local .js files as SERVER_JS

### LINKED ISSUE #10: IDE support? (IntelliJ, VS Code)
Are any of you working with gs files inside IntelliJ by any chance?  Would be really nice to have a plugin to provide support.  I found references to an older plugin that has been removed ( https://github.com/MichaelSnowden/intellij-gas-plugin ) with a clone here: ( https://github.com/kehh/intellij-gas-plugin ), but haven't tried it out yet to see if it works with clasp.

jondcallahan:
fixes #10 

Save `.gs` files with the `.js` extension when fetching project files. Assign type `SERVER_JS` to local `.js` files when pushing to server.

### DIFF
--- a/index.js
+++ b/index.js
@@ -256,8 +256,8 @@ function getAPICredentials(cb) {
  */
 function getFileType(type) {
   return {
-    SERVER_JS: 'gs',
-    SHARED_JS: 'gs'
+    SERVER_JS: 'js',
+    SHARED_JS: 'js'
   }[type] || type.toLowerCase();
 }
 
@@ -270,6 +270,7 @@ function getAPIFileType(path) {
   let extension = path.substr(path.lastIndexOf('.') + 1).toUpperCase();
   return {
     GS: 'SERVER_JS',
+    JS: 'SERVER_JS',
     HTML: 'HTML',
     JSON: 'JSON'
   }[extension];


grant:
Have you tested pushing two files, named `test.gs` and `test.js`? I think we need to handle this case. I think the API would fail if we do that.

What if someone pulls with the old version of clasp, and now pull with this new version of clasp? Won't they get duplicate files (different extensions)?

I think we'll have to break the workflow (users will have to manually rename files), in order to get this right. A potential solution we could implement is renaming all `gs` files to `js` before we push or pull. That way old `gs` files will be upgraded on push or pull.

What do you think?

grant:
I'd say, a reasonable thing to do would be to add some functionality that renames all `.gs` files to `.js` before `push`ing. That way old users don't have a breaking change and all future users will just use `.js`

jondcallahan:
How should we handle name collisions if there are files `Code.js` and `Code.gs` when renaming?

My initial approach of adding 
```javascript
if (path.extname(name).toLowerCase() === '.gs') {
    // Rename local .gs files to .js
    fs.rename(name, `${nameWithoutExt}.js`)
}
```
at line 487 inside `readMultipleFiles` will overwrite `Code.js` with `Code.gs`. Should the tool afford for this case, or blindly rename the `.gs` files to `.js`?

grant:
Let's fail pushing if there is a `.gs` and `.js` file of the same name. So:

- `clasp push`
  - If gs/js conflict:
      - [Prompt](https://www.npmjs.com/package/cli-prompt) the user to **rename** `.gs` to `.js`:
        - ```clasp uses .js files for Apps Script. Rename all .gs files to .js (y/n)?```
        - `y`: For all files in the clasp project, rename `.gs` files to `.js`
        - `n/anything else`: Log "Aborting push"
        - (Some GitHub repos have `.gs` files, we should provide a good experience with clasp still.)
  - If no gs/js conflict:
    - Rename and console.log a message `Rename all .gs files to .js.`
    - Push

- `clasp pull`
  - If gs file found
    - console.log a message saying "Renamed XXX.gs to XXX.js"
    - Pull
  - If no gs/js conflict
    - Pull

`clasp` isn't really smart when pulling anyways (we don't merge, erase, or otherwise sync when pulling)

There are still lots of Apps Script projects with `.gs` like [googlesamples/apps-script](https://github.com/googlesamples/apps-script), and so we need to support `.gs` still.

I think users will have to deal with duplicate files no matter what due to previous versions.

grant:
(Let me know if you want me to work on the logic of https://github.com/google/clasp/pull/27#issuecomment-360651709. I'm going to try to update `clasp` on npm with the latest PRs by Feb 5.)

jondcallahan:
@grant Yes, I would appreciate you working on the logic, thanks. I dove in for a bit today but won't have time to pick it back up for a little while. One open question is how to handle the renaming after a user enters `y`. I think we should make it clear that `Code.js` will be overwritten with the contents of `Code.gs` or create a new file with a name like `Code.gs.js` to preserve the contents of `Code.js`.

DimuDesigns:
I like the idea, but some users may want to retain a *.gs centric workflow. Maybe clasp needs to support a configuration mode where users can opt-in or opt-out of this feature. Maybe something like:
`clasp config [CONFIG-FLAG] [SETTING-FOR-FLAG]`.

grant:
I'm open to PRs that add `extension` to [`.clasp.json`](https://github.com/google/clasp/blob/master/index.ts#L84) and implement custom extensions (`gs`/`js`/`ts`).

Issue #2, opened 2018-01-12
Add an option for login --no-localhost


mhawksey:
Interested in running this in Google Cloud Shell but currently `login` redirects to localhost. A `--no-localhost` option would be useful.

rielzzapps:
If you receive the URL, copy it.

In cloud shell just paste:

wget [url] to call the service and authenticate.

I agree, a --no-localhost would be nice.

Riël Notermans
Dinsdags vrij • Zzapps B.V. • High Tech Campus 9, 5656 AE Eindhoven
<https://www.google.nl/maps/place/Zzapps/@51.4099441,5.4510368,16z/data=!4m5!1m2!3m1!2sZzapps!3m1!1s0x0000000000000000:0x3ad8d7839ef5a5dd>
•
T 040 711 41 94 <+31407114194> • E riel@zzapps.nl • www.zzapps.nl
WIJ MAKEN HET ONDENKBARE MOGELIJK

2018-01-12 15:22 GMT+01:00 mhawksey <notifications@github.com>:

grant:
This is a good idea. Let's think about what the design should be:

```
$ clasp login --no-localhost
> Open: http://console.google.com/...
> Paste the token: ...
> Saved the credentials to ~/.clasp.rc.json.
```

[gcloud's browserless solution: `gcloud auth login --no-launch-browser`](https://cloud.google.com/sdk/gcloud/reference/auth/login)

Is there a common name for browserless login flag (`-f` and/or `--flag`) we should use?

## Temporary workaround:
Copy `~/.clasprc.json` to your machine.

mhawksey:
The command/flag for the Firebase CLI is `login --no-localhost` As this is also Google Auth flow you might be able to reuse some of the code from their implementation https://github.com/firebase/firebase-tools

grant:
The initial `login` command didn't use localhost. When we tackle this feature, we can just use the previous `login` via command line with the code here:

https://github.com/google/clasp/blob/75882839f152a6183c7c92a525069ea6bccecaa8/index.js#L336-L355

grant:
Will close this issue in the next release.

PR #28, opened 2018-01-26
Add headless oauth support

sqrrrl:
Adds --no-localhost option to login command + refactors auth a little.

Planning a subsequent change to upgrade the client lib & auth library to the recently released version (v25.x for client, v1.x) since the new auth library has support for code_verifier in the oauth flow which will makes things more secure.

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,7 +41,6 @@ const Spinner = require('cli-spinner').Spinner;
 const splitLines = require('split-lines');
 const url = require('url');
 const readline = require('readline');
-
 require('http-shutdown').extend();
 
 // Debug
@@ -257,25 +256,22 @@ function getAPICredentials(cb) {
  *     to handle the auth response. False if manual entry used.
  */
  function authorize(useLocalhost) {
-  var opts = {
+  let opts = {
     access_type: 'offline',
     scope: [
       'https://www.googleapis.com/auth/script.deployments',
       'https://www.googleapis.com/auth/script.projects',
     ],
   }
-  var authCode = useLocalhost ?
+  let authCode = useLocalhost ?
     authorizeWithLocalhost(opts) :
     authorizeWithoutLocalhost(opts);
 
   authCode.then(code => {
     return new Promise((res, rej) => {
       oauth2Client.getToken(code, (err, token) => {
-        if (err) {
-          reject(err);
-        } else {
-          res(token);
-        }
+        if (err) return rej(err);
+        return res(token);
       })
     })
   })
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -40,6 +40,9 @@ const recursive = require('recursive-readdir');
 const Spinner = require('cli-spinner').Spinner;
 const splitLines = require('split-lines');
 const url = require('url');
+const readline = require('readline');
+
```
grant: Remove \n.

index.js
```diff
@@ -249,6 +251,100 @@ function getAPICredentials(cb) {
...
+/**
+ * Requests authorization to manage Apps Scrpit projects. Spins up
+ * a temporary HTTP server to handle the auth redirect.
+ *
+ * @param {Object} opts OAuth2 options
+ * @return {Promise} Promise resolving with the authorization code
+ */
+function authorizeWithLocalhost(opts) {
+  return new Promise((resolve, reject) => {
+    var server = http.createServer(function(req, res) {
```
grant: Use `(req, res) =>`.

index.js
```diff
@@ -249,6 +251,100 @@ function getAPICredentials(cb) {
...
+      var url_parts = url.parse(req.url, true);
```
grant: Use `urlParts`. No snake case. Here and elsewhere.

index.js
```diff
@@ -249,6 +251,100 @@ function getAPICredentials(cb) {
...
+    authorizeWithoutLocalhost(opts);
+
+  authCode.then(code => {
+    return new Promise((res, rej) => {
+      oauth2Client.getToken(code, (err, token) => {
+        if (err) {
+          reject(err);
+        } else {
+          res(token);
+        }
```
grant: Let's shorten when possible:

```js
if (err) return reject(err);
return res(token);
```

index.js
```diff
@@ -249,6 +251,100 @@ function getAPICredentials(cb) {
   });
 }
 
+/**
+ * Requests authorization to manage Apps Scrpit projects.
+ * @param {Boolean} useLocalhost True if a local HTTP server should be run
+ *     to handle the auth response. False if manual entry used.
+ */
+ function authorize(useLocalhost) {
+  var opts = {
```
grant: Use `let` instead of `var` everywhere.

index.js
```diff
@@ -249,6 +251,100 @@ function getAPICredentials(cb) {
...
+ * used.
+ *
+ * @param {Object} opts OAuth2 options
+ * @return {Promise} Promise resolving with the authorization code
+ */
+
+function authorizeWithoutLocalhost(opts) {
+  oauth2Client._redirectUri = REDIRECT_URI_OOB;
+  var authUrl = oauth2Client.generateAuthUrl(opts);
+
```
grant: Remove \n

index.js
```diff
@@ -249,6 +251,100 @@ function getAPICredentials(cb) {
...
+  var opts = {
+    access_type: 'offline',
+    scope: [
+      'https://www.googleapis.com/auth/script.deployments',
+      'https://www.googleapis.com/auth/script.projects',
+    ],
+  }
+  var authCode = useLocalhost ?
+    authorizeWithLocalhost(opts) :
+    authorizeWithoutLocalhost(opts);
```
grant: Great ternary!

index.js
```diff
@@ -249,6 +251,100 @@ function getAPICredentials(cb) {
...
+      if (url_parts.query.code) {
+        resolve(url_parts.query.code);
+      } else {
+        reject(url_parts.query.error);
+      }
+      res.end(LOG.AUTH_PAGE_SUCCESSFUL);
+      server.shutdown();
+    }).withShutdown();
+
+    server.listen(0, () => {
```
grant: Listen on port 0? I assume it's not in use. o.w. we'll get `ERR_SERVER_ALREADY_LISTEN`.

grant:
Fixes #2.

index.js
sqrrrl: 0 just means pick an unused ephemeral port. For installed app flows, port is ignored for registered callback localhost URLs, specifically for this reason -- hardcoded ports can cause conflicts if there's already a service running on it.

index.js
sqrrrl: done

index.js
sqrrrl: done

index.js
sqrrrl: done

index.js
sqrrrl: done

index.js
sqrrrl: done

sqrrrl:
PTAL

PR #40, opened 2018-02-08
Added the option to clone a specific version of a script.

diminishedprime:
I found myself needing to sometimes clone a specific version of a script, and this addition makes it possible to do that via clasp.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -490,12 +490,14 @@ commander
  * `pwd` with dots converted to subdirectories.
  * @param {string} scriptId The project script id
  * @param {string?} rootDir The directory to save the project files to. Defaults to `pwd`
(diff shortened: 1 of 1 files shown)


grant:
Awesome, thanks!

grant:
I think this PR forgot to add the version number to the `fetchProject` call.

diminishedprime:
Hmm. I wasn't sure about fetchProject when I was working on this. I was thinking of versions as immutable, so I thought subsequent fetches wouldn't need to know the version. I guess versions are allowed to change, though, right?

erickoledadevrel:
I think Grant meant that although the clone command takes a `versionNumber` parameter, you aren't using it anywhere.

Issue #41, opened 2018-02-09
Offline Error Messages


labnol:
If a user executes the clasp push command while the network is silently down, the following errors are shown.  Can the script check for network connection before proceeding?

```
Push failed. Errors:
(node:67970) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'map' of undefined
    at /Users/labnol/digitalinspiration/node_modules/@google/clasp/index.js:587:46
    at /Users/labnol/digitalinspiration/node_modules/google-auth-library/build/src/auth/oauth2client.js:396:24
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:160:7)
(node:67970) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:67970) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

grant:
To handle this, we could add an [`is-online`](https://github.com/sindresorhus/is-online) check before every network call.

arjun-rao:
@grant I guess this issue can be closed now.

PR #47, opened 2018-02-21
FIX: Gracefully handle network errors

arjun-rao:
Implemented [is-online](https://github.com/sindresorhus/is-online) check for cases where there is a network call to solve #41

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -44,6 +44,8 @@ const readline = require('readline');
 import * as Promise from 'bluebird';
 import { Server } from "http";
 
+const isOnline = require('is-online');
+
 // Debug
 const DEBUG = false;
 
@@ -158,6 +160,7 @@ const LOG = {
   PULLING: 'Pulling files...',
   PUSH_SUCCESS: (numFiles: number) => `Pushed ${numFiles} ${pluralize('files', numFiles)}.`,
   PUSH_FAILURE: 'Push failed. Errors:',
+  NO_NETWORK: 'Error: Looks like you are not connected to the internet',
   PUSHING: 'Pushing files...',
   REDEPLOY_END: 'Updated deployment.',
   REDEPLOY_START: 'Updating deployment...',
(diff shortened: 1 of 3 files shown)


index.ts
```diff
@@ -44,6 +44,8 @@ const readline = require('readline');
 import * as Promise from 'bluebird';
 import { Server } from "http";
 
+const isOnline = require('is-online');
```
grant: Please put this in alphabetical order above.

index.ts
```diff
@@ -44,6 +44,8 @@ const readline = require('readline');
...
+
```
grant: No \n

index.ts
```diff
@@ -158,6 +160,7 @@ const LOG = {
   PULLING: 'Pulling files...',
   PUSH_SUCCESS: (numFiles: number) => `Pushed ${numFiles} ${pluralize('files', numFiles)}.`,
   PUSH_FAILURE: 'Push failed. Errors:',
+  NO_NETWORK: 'Error: Looks like you are not connected to the internet',
```
grant: - Please put this message under `ERROR`
- `NO_NETWORK` -> `OFFLINE`
- Something like`Error: Looks like you are offline.`
  - Period to end sentence
  - Be terse

index.ts
```diff
@@ -273,11 +276,17 @@ function getProjectSettings(): Promise<ProjectSettings> {
...
-  DOTFILE.RC.read().then((rc: ClaspSettings) => {
-    oauth2Client.credentials = rc;
-    cb(rc);
-  }).catch((err: object) => {
-    logError(null, ERROR.LOGGED_OUT);
+  isOnline().then((online:boolean)=>{
+    if (!online){
+      console.error(LOG.NO_NETWORK);
+      process.exit(1);
+    }
```
grant: This is repeated many times.
```js
      if (!online){
        console.error(LOG.NO_NETWORK);
        process.exit(1);
      }
```
Make it a 1-line function.
```js
// usage
checkIfOnline();

// fn
function checkIfOnline() {
  if (!(await isOnline()) {
    logError(null, ERROR.OFFLINE));
    process.exit(1);
  }
}
```

index.ts
```diff
@@ -465,23 +480,23 @@ commander
     if (fs.existsSync(DOT.PROJECT.PATH)) {
       logError(null, ERROR.FOLDER_EXISTS);
     } else {
-      getAPICredentials(() => {
```
grant: This looks like 7 columns.

arjun-rao:
@grant  PTAL: I've updated the PR with the changes requested.

index.ts
```diff
@@ -613,9 +631,9 @@ commander
               if (abortPush) return spinner.stop(true);
 
               const files = filePaths.map((name, i) => {
-                let nameWithoutExt = name.slice(0, -path.extname(name).length);
-                // Formats rootDir/appsscript.json to appsscript.json. 
-                // Preserves subdirectory names in rootDir 
+                const nameWithoutExt = name.slice(0, -path.extname(name).length);
```
grant: This change will conflict with #48. Don't change code that's out of scope of the fix next time.

appliedcs: Will keep that in mind. npm lint complained that there was let being instead of const on those two lines so I just changed it. I'll scope my PRs better next time.

arjun-rao: I got an error on  tslint when let was used in those lines so I changed them to const. I'll keep this in mind for future PRs.

Issue #12, opened 2018-01-17
Support options to specify dotfiles


fossamagna:
Please support options to specify dotfiles. e.g: .clasp.json, .clasprc.json.

I would like to switch enviroment to deploy. eg: development, production...
So, If I can specify dotfiles by clasp's options, it is very useful.

grant:
Since it's pretty easy just to write a bash command to swap/move `.clasp.json` files, I'm probably not going to think about this for a bit.

You could just have:
```
.clasp.prod.json
.clasp.dev.json
```
And `cp` which one you need:
```
cp .clasp.prod.json .clasp.json
```

JeanRemiDelteil:
@fossamagna you can use a GAS starter project with a build that will swtich those files for you ([https://github.com/JeanRemiDelteil/gas-shell](url))
This project also comes with a number possible configurations to patch the gas manifest file for example.

dlindahl:
I'm not hear to start a war or anything, but do realize that not everyone knows how to write bash scripts. CLI arguments are, in my opinion, much more approachable to a wider audience. I also just happen to think that it makes for a cleaner, declarative solution to working with environment-specific configuration needs.

JeanRemiDelteil:
Other issue is that often bash script differ from OS.
When you work with a team it's difficult to get this working everywhere.

grant:
I support this feature, a more detailed potential solution would be nice.

I propose the .clasp.json file should have an env property like npm:

https://docs.npmjs.com/misc/config

A more detailed spec of what this would look like would be important.

PRs are welcome.

fossamagna:
@grant 
I think that use-cases is the following:

- I want to push to different scriptId. For example, switch environment as development, staging, production. Many users use application on production environment. I want to develop (coding, test, etc) on development environment. I pull source code from development environment and push staging or production environment.
-  I use mutilple Google accounts on different domains. I develop a application on one domain and then I deploy the application other domain.  I need to use credential for different domain at same time.

grant:
OK. There are 2 separate requests, **multi-scriptIds** & **multi-account**.
This is complicated, bear with me..

## Multiple `scriptId`s for the same `clasp` project (`ProjectSettings`)
**Question:** Why does someone need to push to multiple parallel Apps Script projects, like dev/prod?
- 1 clasp project can push to multiple `scriptId`s (dev/staging/prod).
  - Currently all API requests use: `.clasp.json` (`scriptId`/`rootDir`)
[...]

PR #56, opened 2018-03-05
Add env option to swicth environment.

fossamagna:
I changed to contains multiple environments(scriptId) in `.clasp.json`.
Environment is specify by `env` option. If `env` option isn't specify, used as environment name is `default`.

refs #12

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -84,7 +84,7 @@ interface ClaspSettings {
 }
 // Project settings file (Saved in .clasp.json)
 interface ClaspProjectSettings {
-  [index: string]: ProjectSettings
+  [index: string]: ProjectSettings;
 }
 
 interface ProjectSettings {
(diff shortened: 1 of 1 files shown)


fossamagna:
I signed it!

index.ts
```diff
@@ -97,6 +103,50 @@ interface LoginOptions {
   localhost: boolean;
 }
 
+interface CreateOptions {
+  env: string;
```
grant: We can't just add env options to every command. They need to be in `.clasp.json`.
https://github.com/google/clasp/issues/12#issuecomment-370672417

index.ts
```diff
@@ -81,6 +83,10 @@ interface ClaspSettings {
   expiry_date: string;
 }
 // Project settings file (Saved in .clasp.json)
+interface ClaspProjectSettings {
```
grant: `ClaspProjectSettings` is not needed. How is it different than `ProjectSettings`?

fossamagna: `ClaspProjectSettings` is object to enclose `ProjectSettings` and has multiple `ProjectSettings`.
In https://github.com/google/clasp/issues/12#issuecomment-370672417, `ClaspProjectSettings` equivalent to `ProjectSettings` and `ProjectSettings` equivalent to `Env`.

index.ts
fossamagna: I think that `env` option is need to specify envId in `.clasp.json`. If no `env` option, How do it specify `envId` from CLI ?

grant: I don't support a --env flag in each CLI command. Follow https://github.com/google/clasp/issues/12#issuecomment-370672417 carefully. It's very complex and I don't know if I covered all the cases. We should probably discuss more before implementing.

index.ts
grant: Can you use the exact terms as in the comment? I wrote:

```ts
// Project settings file (Saved in .clasp.json)
interface ProjectSettings {
  scriptId: string; // default env
  rootDir: string;  // default env
  envs: Map<EnvId, Env> // custom scriptIds (NEW)
}
```

`ClaspProjectSettings` is redundant. Either say `ClaspSettings` or `ProjectSettings`.

fossamagna: I see. I will use extact terms.

index.ts
fossamagna: I understood your idea.
I have a consideration for use-case of multiple google accounts and multiple `scriptId`s at same time. Please see https://github.com/google/clasp/issues/12#issuecomment-371168118

grant:
I don't think this PR is getting closer to adding environments to clasp projects. Do you mind if we close this PR and discuss more in #12?

PR #60, opened 2018-03-08
Fix replacing only the first '\' character in path

JeanRemiDelteil:
Current implementation only replace the first '\' char,
this will replace all '\\' chars in the file path

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -633,7 +633,7 @@ commander
               const files = filePaths.map((name, i) => {
                 let nameWithoutExt = name.slice(0, -path.extname(name).length);
                 // Replace OS specific path separator to common '/' char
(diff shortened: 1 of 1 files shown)


grant:
Whoops, thanks!

HeavensBlade:
I also realized that behavior.
When I created a nested structure the filename looks like this: folder/subfolder\file.html
When including I needed to escape the backslash: <?!= include('folder/subfolder\\file.html'); ?>

Would I need to change that in all my projects when this fix is implemented?

JeanRemiDelteil:
You will need to rename calls to templates when the fix is applied, yes.



Jean-Rémi Delteil,
Awesome Table team

PR #61, opened 2018-03-09
Document available configuration in README.md

aurelienshz:
This PR adds a configuration reference to the project's README. The idea behind this is that the ability to use a different rootDir is worth mentioning without having to read the code (I had overlooked it at first : https://github.com/danthareja/node-google-apps-script/issues/86#issuecomment-371725779)


grant:
Thanks. We should probably shorten this documentation to make it more readable, but this is a good start.

Issue #38, opened 2018-02-05
Allow manifest, file list via CLI


sqrrrl:
Currently the file set to upload in the push command is automatically determined. It would be helpful to be able to specify files (both manifest and js) via the command line as well.

The particular use case is to be able to deploy different configurations -- e.g. in test mode I want to use an alternate manifest with execution API enabled, tests bundled, etc. while in prod mode I want to skip all that. This is particularly important for libraries where we don't want to have long-lived deployments, but may want to temporarily deploy as executable or web app for purposes of running tests and don't want to package code that is only used for resting.. Same can be used for add-ons and other things where we'd like to keep test code & configuration somewhat isolated.

My ideal is to have commands in package.json like:

"scripts": {
  "push-test": "clasp push --env=test appsscript.test.json test/**.js src/**.js"
  "push-prod": "clasp push --env=prod appscript.prod.json src/**.js",
  ...
}

Many CLIs allow specifying files via command line -- tsc, babel, jshint, etc.. all allow specifying which files to operate on as args. Typescript is also very close to what is being asked for. By default, it'll pick up all *.ts files as is the current behavior of clasp. But that can be overridden by explicitly providing a list of globs as args.

grant:
Do you mean test/prod options like #12?

sqrrrl:
Sort of, but not just the project ID. Suppose the same technique could be used with renaming .claspignore and manifest files, but it's clunky. Better would be to just take the file patterns as optional arguments. If they're present, push those files. If not, fall back on the current algorithm.

Ideally we'd have both an environment option for auth/project ID *and* a way to easily change file sets. I'd like my package.json to look something like:

"scripts": {
  "push-test": "clasp push --env=test appsscript.test.json test/**.js src/**.js"
  "push-prod": "clasp push --env=prod appscript.prod.json src/**.js",
  ...
}

grant:
I don't see a use-case where a 1-line bash script can't solve. I'd rather have the CLI be minimal and use `git`'s features or a bash file instead.

sqrrrl:
A few reasons:

*  It's not a 1 line script. Also need to make sure the ephemeral files are ignored by git or whatever source control system is being used.
* Sometimes expressing things in the affirmative is easier than expressing them in the negative. Right now clasp only allows expressing things negatively -- tell me the files you *don't* want to push. In more complex projects, the affirmative expression can end up being more concise.
[...]

PR #62, opened 2018-03-10
WIP: Use commands programmatically

aurelienshz:
## Purpose:
Make all CLI features usable prgrammatically, in a Node.js script. This helps for some advanced use cases where the users want to be able to script the upload process, without making the CLI more complex. For example #38 could be solved by having an upload script responsible for reading the environment config files, and write the correct config values before calling clasp. This also allows to have Clasp integrate nicely with a build system such as Gulp.

Another nice side effect is that this PR makes testing possible (since we can import the features from a test file).

## Updates performed:

* Make the build script of this tool capable of handling several files inside a src directory
* Import these features in cli.ts, where all things CLI-related have been moved (ie all commander calls)


grant:
@aurelienshz Thanks for the PR.
I don't think I can approve of a PR that modifies the whole library without having a large discussion and design doc around it.
### Do you mind
- Discussing what issue you face
- Propose a small PR that fixes that issue?

I think we should this PR in favor of discussing first and making small PRs.

aurelienshz:
Hi @grant !

You are perfectly right. I started refactoring and got completely carried away, and started over-refactoring and writing the PR draft alongside with random ideas that came to me. Sorry :blush: 

I have just updated the description and pushed commits on the branch that make the scope of the PR a lot clearer, and greatly reduce the changeset.

src/cli.ts
```diff
@@ -0,0 +1,99 @@
...
+import {
+  PROJECT_NAME, ERROR, LoginOptions,
+  login, logout, create, clone, pull, push, openScriptProject,
+  listDeployments, deploy, undeploy, redeploy, listVersions, createVersion
+} from "./index";
+
+/**
+ * Set global CLI configurations
+ */
+commander
```
grant: Laying out `commander` commands like this is nice.
I'd support this change.

grant:
I appreciate the enthusiasm, but I don't think there are many advantages of using clasp as an API. There's the Apps Script API for that, and it's not complex.

This PR is changing too much. If you want to modify this PR to **only** move the `command` commands together (like I commented) and not export the commands, I'd approve of that. 

Otherwise, I'm in favor of closing this and discussing more in GitHub issues.

aurelienshz:
Closed in favor of #68

grant:
If we want to create an API, we need smaller PRs and more discussion about design and code quality.
I don't see why you can't build an API that calls the CLI if that's the goal.

PR #71, opened 2018-03-21
Add versionNumber as parameter to clone

diminishedprime:
My previous PR (#40) didn't use the versionNumber argument to clasp clone, so added that in now.


grant:
Thanks.

Issue #80, opened 2018-03-30
tsconfig.json not used in tsc


grant:
`tsconfig.json` not used by `tsc`.

Reported by @erickoledadevrel

PR #91, opened 2018-04-04
Use tsconfig file

campionfellin:
According to: https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

> When input files are specified on the command line, tsconfig.json files are ignored.

This fixes https://github.com/google/clasp/issues/80


grant:
Thanks!

Issue #69, opened 2018-03-19
List Apps Script Projects (`clasp list`)


grant:
`clasp list`

Does a Drive search for Apps Script projects using `files.list`.

https://developers.google.com/drive/v3/reference/files/list
`q`:`mimeType="application/vnd.google-apps.script"`

PR #93, opened 2018-04-04
Create clasp list function

campionfellin:
This creates the clasp list function as per https://github.com/google/clasp/issues/69

You can use it by running `clasp list`

In the future, it would be nice to have it highlight your current project.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -156,6 +156,7 @@ const LOG = {
   DEPLOYMENT_LIST: (scriptId: string) => `Listing deployments for ${scriptId}...`,
   DEPLOYMENT_START: (scriptId: string) => `Deploying project ${scriptId}...`,
   FILES_TO_PUSH: 'Files to push were:',
+  FINDING_SCRIPTS: 'Looking for your scripts now...',
   OPEN_PROJECT: (scriptId: string) => `Opening script: ${scriptId}`,
   PULLING: 'Pulling files...',
   PUSH_SUCCESS: (numFiles: number) => `Pushed ${numFiles} ${pluralize('files', numFiles)}.`,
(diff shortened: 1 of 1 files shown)


index.ts
```diff
@@ -931,6 +933,41 @@ commander
     });
   });
 
+/**
+ * Lists the apps scripts
```
grant: Can you give more detail example? Answer these questions:
- Does it return all my scripts? (no)
- What does the output look like (name + id) (Give an `@example`)

index.ts
```diff
@@ -156,6 +156,7 @@ const LOG = {
   DEPLOYMENT_LIST: (scriptId: string) => `Listing deployments for ${scriptId}...`,
   DEPLOYMENT_START: (scriptId: string) => `Deploying project ${scriptId}...`,
   FILES_TO_PUSH: 'Files to push were:',
+  FINDING_SCRIPTS: 'Looking for your scripts now...',
```
grant: Pretty good, but I think we can be more terse.
Can you write:
'Finding your scripts...'

index.ts
```diff
@@ -931,6 +933,41 @@ commander
...
+    getAPICredentials(async () => {
+      await checkIfOnline();
+      getProjectSettings().then(({ scriptId }: ProjectSettings) => {
+      const drive = google.drive({version: 'v3', auth: oauth2Client});
+      drive.files.list({
+        pageSize: 10,
+        fields: 'nextPageToken, files(id, name)',
+        q: "mimeType='application/vnd.google-apps.script'",
+      }, function(err: any, { data }: any) {
+        spinner.stop(true);
```
grant: Arrow function:
```js
}, (err: any, { data }: any) => {
```

index.ts
```diff
@@ -931,6 +933,41 @@ commander
...
+        if (err) {
+          return console.error(ERROR.UNAUTHENTICATED);
+        }
```
grant: If you want, for 1 line returns, you can write:
```js
if (err) return console.error(ERROR.UNAUTHENTICATED);
```

Issue #46, opened 2018-02-21
View Stackdriver logs


grant:
## Expected Behavior

```sh
clasp logs
(DEBUG) {foo:42} – myFunction [2018-02-21T16:08:06.364Z]
```

```sh
clasp logs --json
[{
  "jsonPayload": {foo:42},
  ...
}, {...}]
```

Views StackDriver logs (`console.log`).

## Actual Behavior

Command doesn't exist yet.

## Notes

- [Stackdriver API (entries/list)](https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/list)
- [Apps Script console.log](https://developers.google.com/apps-script/reference/base/console#methods)

### Example Stackdriver API result

```json
{
  "insertId": "13oj0ajfyb6tdx",
  "jsonPayload": {foo:42},
  "labels": {…},
  "logName": "projects/project-id-4088669541713651222/logs/script.googleapis.com%2Fconsole_logs",
  "receiveTimestamp": "2018-02-21T16:08:07.370458993Z",
  "resource": {
    "labels": {
      "function_name": "myFunction",
      "invocation_type": "editor",
      "project_id": "project-id-4088669541713651222",
    },
    "type": "app_script_function"
  },
  "severity": "DEBUG",
  "timestamp": "2018-02-21T16:08:06.364Z",
}
```

Please 👍 if you'd like this feature.

campionfellin:
I'd like to take this on!
I have written a basic logs function, which you can see here:

https://github.com/campionfellin/clasp/commit/01b0b42623ef9784874cd7daade86587ab151fd3

It needs some cleanup and styling, but should work.

grant:
Sweet! Looking good!

1 nit. We'll probably want to use the `ProjectSettings` instead of an env variable. Like:
```ts
// Project settings file (Saved in .clasp.json)
interface ProjectSettings {
  scriptId: string;
  rootDir: string;
  projectId: string; // Resources -> Cloud Platform Project...
}
```

campionfellin:
Yes, I saw the ProjectSettings, and I wasn't sure how we wanted to implement actually getting the projectId.

I see a couple options:

1) Make a command like `clasp config set projectId=test123`
       This would likely mean making other subcommands under `clasp config`
2) Run `gcloud projects list` for the user, and allow them to choose the project they have associated with the Google Script.

[...]

PR #98, opened 2018-04-06
Do Not Merge:  clasp logs

campionfellin:
This is a WIP, it gets the logs and everything, we just need to decide how we want to print out the logs... I don't think that with

```
console.log("=================");
console.log(entries[i].data);
console.log("=================");
```

is ideal. Do we want to format it exactly like in https://github.com/google/clasp/issues/46 ?

Additionally, it might be nice to use something like https://github.com/Marak/colors.js to distinguish different types of log messages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,21 @@ clasp deployments # List all deployment IDs
 clasp open
 ```
 
+### See CloudDriver logs from script
+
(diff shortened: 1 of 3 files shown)


campionfellin:
Sweet, I will simplify it and add those subcommands. Probably will update it on Saturday.

campionfellin:
Closing in favor of https://github.com/google/clasp/pull/101

Issue #97, opened 2018-04-05
Should throw error when running `clasp create` if there is a `.clasp.json` in a parent directory. 


diminishedprime:
## Expected Behavior
When running `clasp create` in a child directory of a clasp project, clasp should throw an error explaining that nested projects are not supported. 

## Actual Behavior
Clasp silently uses the `.clasp.json` file of the parent directory, but creates a new project at the `process.cwd()`

## Steps to Reproduce the Problem

  1. create a clasp project in directory `a`
  1. `cd` into `a` and create a new sub-directory `b`
  1. run clasp create again with a new project name
  1. note that there is no `.clasp.json` file created in directory b.

## Specifications

  - Node version (`node -v`)
    - v9.10.1
  - Version (`npm list -g | grep clasp`):
    - latest from using npx
  - OS (Mac/Linux/Windows)
    - linux

diminishedprime:
Likely related to issue #92

grant:
Solution:
- `clasp create` should `getProjectSettings`. [Here](https://github.com/google/clasp/blob/master/index.ts#L579)
  - if getProjectSettings promise resolves, then log an error saying nested `clasp` projects are not supported.

PRs welcome.

campionfellin:
This was fixed in #99

PR #99, opened 2018-04-06
clasp create now checks for .clasp.json files

campionfellin:
As per https://github.com/google/clasp/issues/97 when you run `clasp create` in a directory with a `.clasp.json` file or in a directory under any other directory that has a `.clasp.json` file, then you get an error saying `Nested clasp projects are not supported.`

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -250,15 +250,18 @@ const getScriptURL = (scriptId: string) => `https://script.google.com/d/${script
 /**
  * Gets the project settings from the project dotfile. Logs errors.
  * Should be used instead of `DOTFILE.PROJECT().read()`
+ * @param  {boolean} failSilently if you don't want to err when it doesn't
+ * find a dot file (such as when creating a a new script)
  * @return {Promise} A promise to get the project script ID.
  */
-function getProjectSettings(whileCreating): Promise<ProjectSettings> {
+function getProjectSettings(failSilently: boolean): Promise<ProjectSettings> {
   const promise = new Promise<ProjectSettings>((resolve, reject) => {
-    const fail = (whileCreating) => {
-      if (!whileCreating) {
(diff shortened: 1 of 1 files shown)


index.ts
```diff
@@ -251,10 +252,12 @@ const getScriptURL = (scriptId: string) => `https://script.google.com/d/${script
  * Should be used instead of `DOTFILE.PROJECT().read()`
  * @return {Promise} A promise to get the project script ID.
  */
-function getProjectSettings(): Promise<ProjectSettings> {
+function getProjectSettings(whileCreating): Promise<ProjectSettings> {
```
grant: - Rename to `failSilently`.
- Add this `@param`

index.ts
```diff
@@ -251,10 +252,12 @@ const getScriptURL = (scriptId: string) => `https://script.google.com/d/${script
...
   const promise = new Promise<ProjectSettings>((resolve, reject) => {
-    const fail = () => {
-      logError(null, ERROR.SCRIPT_ID_DNE);
+    const fail = (whileCreating) => {
+      if (!whileCreating) {
```
grant:
```ts
if (!failSilently) {
  logError(null, ERROR.SCRIPT_ID_DNE);
  resolve();
}
```

index.ts
```diff
@@ -581,6 +584,14 @@ commander
         await checkIfOnline();
         spinner.setSpinnerTitle(LOG.CREATE_PROJECT_START(title));
         spinner.start();
+        getProjectSettings(true).then(({ scriptId }: ProjectSettings) => {
+          if (scriptId) {
+            console.error(ERROR.NO_NESTED_PROJECTS);
+            process.exit(1);
+          }
+        }).catch((err) => {
```
grant: If you `resolve` the Promise, then you can remove this `catch` handler.

PR #104, opened 2018-04-11
Create clasp run V1 command

campionfellin:
This is what we can call V1 of `clasp run`. There are a couple of things to note for this to work:

You will need to put your `clientId` and `clientSecret` here: https://github.com/google/clasp/blob/57adf76adda47d155fa818d20cc3d88fb15aac8e/index.ts#L135

Then build the tool `sudo npm run build`

Then, use `clasp login --localkey` to login and authorize (saves `.clasprc.json` to `./` instead of `~/.`).

Then go ahead and use `clasp run <scriptId> <theFunction>`

Probably needs a bit of work, but we'll see how helpful this is for now.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -291,21 +291,13 @@ function getProjectSettings(failSilently?: boolean): Promise<ProjectSettings> {
  * @param {Function} cb The callback
  */
 function getAPICredentials(isLocal: boolean, cb: (rc: ClaspSettings | void) => void) {
-  if (isLocal) {
-    DOTFILE.RC_LOCAL.read().then((rc: ClaspSettings) => {
+  const dotfile = isLocal ? DOTFILE.RC_LOCAL : DOTFILE.RC;
+  dotfile.read().then((rc: ClaspSettings) => {
       oauth2Client.setCredentials(rc);
       cb(rc);
     }).catch((err: object) => {
       process.exit(-1);
     });
-  } else {
-    DOTFILE.RC.read().then((rc: ClaspSettings) => {
-      oauth2Client.setCredentials(rc);
-      cb(rc);
-    }).catch((err: object) => {
-      process.exit(-1);
-    });
-  }
 }
 
 /**
@@ -549,17 +541,17 @@ commander
   .command('login')
   .description('Log in to script.google.com')
   .option('--no-localhost', 'Do not run a local server, manually enter code instead')
-  .option('--localkey', 'Save .clasprc.json file locally')
+  .option('--ownkey', 'Save .clasprc.json file to current working directory')
   .action((options: {
     localhost: boolean;
-    localkey: boolean;
+    ownkey: boolean;
   }) => {
     // Try to read the RC file.
     DOTFILE.RC.read().then((rc: ClaspSettings) => {
       console.warn(ERROR.LOGGED_IN);
     }).catch(async (err: string) => {
       await checkIfOnline();
-      authorize(options.localhost, options.localkey);
+      authorize(options.localhost, options.ownkey);
     });
   });
 
@@ -1111,29 +1103,31 @@ commander
   });
 
 /**
- * Clasp run <function>
- * 
- * @param scriptId ID of the script you want to run
+ * Clasp run <functionName>
+ * This function runs your script in the cloud. You must supply
+ * the functionName params. For now, it can 
+ * only run functions that do not require other authorization.
  * @param theFunction function in the script that you want to run
- * 
- * Note: to use this command, you must have used `clasp login --localkey`
+ * @see https://developers.google.com/apps-script/api/reference/rest/v1/scripts/run
+ * Note: to use this command, you must have used `clasp login --ownkey`
  */
 commander
-  .command('run <scriptId> <theFunction>')
-  .description('Run a command')
-  .action((scriptId, theFunction) => {
+  .command('run <functionName>')
+  .description('Run a function in your Apps Scripts project')
+  .action((functionName) => {
     getAPICredentials(true, async () => {
       await checkIfOnline();
-      const params = {
-        'scriptId': scriptId,
-        'function': theFunction,
-        'devMode': false
-      };
+      getProjectSettings().then(({ scriptId }: ProjectSettings) => {
+        const params = {
+          'scriptId': scriptId,
+          'function': functionName,
+          'devMode': false
+        };
 
-      script.scripts.run(params).then(response => {
-        console.log(response.data);
+        script.scripts.run(params).then(response => {
+          console.log(response.data);
+        });
       });
-
     });
   });
 
--- a/index.ts
+++ b/index.ts
@@ -335,7 +335,7 @@ function authorize(useLocalhost: boolean, writeToLocalKey: boolean) {
       oauth2Client.getToken(code).then((token) => res(token.tokens));
     });
   }).then((token: object) => {
-    writeToLocalKey ? DOTFILE.RC_LOCAL.write(token) : DOTFILE.RC.write(token)
+    writeToLocalKey ? DOTFILE.RC_LOCAL.write(token) : DOTFILE.RC.write(token);
   })
     .then(() => console.log(LOG.AUTH_SUCCESSFUL))
     .catch((err: string) => console.error(ERROR.ACCESS_TOKEN + err));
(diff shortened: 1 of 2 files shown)


index.ts
```diff
@@ -288,29 +290,39 @@ function getProjectSettings(failSilently?: boolean): Promise<ProjectSettings> {
...
  */
-function getAPICredentials(cb: (rc: ClaspSettings | void) => void) {
-  DOTFILE.RC.read().then((rc: ClaspSettings) => {
-    oauth2Client.setCredentials(rc);
-    cb(rc);
-  }).catch((err: object) => {
-    process.exit(-1);
-  });
+function getAPICredentials(isLocal: boolean, cb: (rc: ClaspSettings | void) => void) {
+  if (isLocal) {
```
grant:
```ts
const dotfile = isLocal ? DOTFILE.RC_LOCAL : DOTFILE.RC;
dotfile.read().then...
```

index.ts
```diff
@@ -288,29 +290,39 @@ function getProjectSettings(failSilently?: boolean): Promise<ProjectSettings> {
...
+  }
 }
 
 /**
  * Requests authorization to manage Apps Script projects.
  * @param {boolean} useLocalhost True if a local HTTP server should be run
  *     to handle the auth response. False if manual entry used.
  */
-function authorize(useLocalhost: boolean) {
+function authorize(useLocalhost: boolean, writeToLocalKey: boolean) {
```
grant: For every `local`, say `own`.
`writeToOwnKey`.

The global key is also local.

index.ts
```diff
@@ -1094,6 +1110,33 @@ commander
     });
   });
 
+/**
+ * Clasp run <function>
+ *
```
grant: - Describe what this can and can't do. i.e. does it accept params?
- Remove newlines.

index.ts
```diff
@@ -1094,6 +1110,33 @@ commander
...
+/**
+ * Clasp run <function>
+ * 
+ * @param scriptId ID of the script you want to run
+ * @param theFunction function in the script that you want to run
+ * 
+ * Note: to use this command, you must have used `clasp login --localkey`
+ */
+commander
+  .command('run <scriptId> <theFunction>')
```
grant: - Always default to the current script.
  - `.command('run <theFunction>')`
- Use `functionName` instead of `theFunction`

index.ts
```diff
@@ -1094,6 +1110,33 @@ commander
...
+  .description('Run a command')
```
grant: Run a function in your Apps Script project.

index.ts
```diff
@@ -1094,6 +1110,33 @@ commander
...
+  .action((scriptId, theFunction) => {
+    getAPICredentials(true, async () => {
+      await checkIfOnline();
+      const params = {
+        'scriptId': scriptId,
+        'function': theFunction,
+        'devMode': false
```
grant:
```ts
const params {
  scriptId,
  'function': functionName,
  devMode: true,
}
```

index.ts
```diff
@@ -1094,6 +1110,33 @@ commander
     });
   });
 
+/**
+ * Clasp run <function>
+ * 
+ * @param scriptId ID of the script you want to run
+ * @param theFunction function in the script that you want to run
+ *
```
grant: `@see https://developers.google.com/apps-script/api/reference/rest/v1/scripts/run`

index.ts
```diff
@@ -288,29 +290,31 @@ function getProjectSettings(failSilently?: boolean): Promise<ProjectSettings> {
...
-  DOTFILE.RC.read().then((rc: ClaspSettings) => {
-    oauth2Client.setCredentials(rc);
-    cb(rc);
-  }).catch((err: object) => {
-    process.exit(-1);
-  });
+function getAPICredentials(isLocal: boolean, cb: (rc: ClaspSettings | void) => void) {
+  const dotfile = isLocal ? DOTFILE.RC_LOCAL : DOTFILE.RC;
+  dotfile.read().then((rc: ClaspSettings) => {
+      oauth2Client.setCredentials(rc);
```
grant: Spacing.

index.ts
```diff
@@ -1094,6 +1102,35 @@ commander
...
+  .description('Run a function in your Apps Scripts project')
+  .action((functionName) => {
+    getAPICredentials(true, async () => {
+      await checkIfOnline();
+      getProjectSettings().then(({ scriptId }: ProjectSettings) => {
+        const params = {
+          'scriptId': scriptId,
+          'function': functionName,
+          'devMode': true
+        };
```
grant: Minimize `params` (no quotes in keys, make key/value the same if possible).

campionfellin: Got rid of the quotes and of `scriptId`. However, it still appeared to need both `function` and `devMode`

campionfellin:
Should be ready now. Missed a couple of the comments previously, but it looks like I got them all. Let me know if I missed any.

index.ts
```diff
@@ -64,6 +64,7 @@ const DOT = {
   },
   RC: { // Saves global information, in the $HOME directory
     DIR: '~',
+    LOCAL_DIR: './',
```
grant: I think `/` may break Windows.
Then again, I think there are other problems with `clasp` on Windows.

index.ts
```diff
@@ -1094,6 +1102,35 @@ commander
...
+ * Note: to use this command, you must have used `clasp login --ownkey`
+ */
+commander
+  .command('run <functionName>')
+  .description('Run a function in your Apps Scripts project')
+  .action((functionName) => {
+    getAPICredentials(true, async () => {
+      await checkIfOnline();
+      getProjectSettings().then(({ scriptId }: ProjectSettings) => {
+        const params = {
```
grant: This could be inlined.

package.json
```diff
@@ -29,6 +29,7 @@
   "dependencies": {
     "@google-cloud/logging": "^1.2.0",
     "anymatch": "^1.3.2",
+    "axios": "^0.18.0",
```
grant: Don't think these are needed, but maybe I forgot something.

campionfellin: Yeah it may just be my machine, but I can't seem to build without `axios` and `grpc` 😔

jasonjurotich:
Getting errors when trying clasp run. See txt.
[Untitled document.txt](https://github.com/google/clasp/files/2058515/Untitled.document.txt)

campionfellin:
Hey @jasonjurotich, can you show us your `myFunction` code?

jasonjurotich:
function myFunction() {
SpreadsheetApp.openById('1acJRcIGXIYa9YNgO8g3DTrgFRO6Iqk0wNJoxYpsust8').getRange('A1').setValue('etset');
}

It was just an experiment. I even ran it like normal before going back to clasp to try it so I know it works fine on its own.

grant:
That's a 404. Not sure why. Maybe the deployment is not set up properly or something else is going wrong.

campionfellin:
Here's a couple of quick thoughts:

Did you deploy through the UI?

`clasp open`
**Publish** --> **Deploy as web app...**

When doing this, I get 

```
  errors:
   [ { message: 'The caller does not have permission',
       domain: 'global',
       reason: 'forbidden' } ] }
``` 

which I think means I need to change the permissions on the Sheet that I'm trying to access. This could also have to do with the scopes we give https://github.com/google/clasp/blob/2d8ba8ee71edb2e84bfbfa10c601844d09d7212b/src/auth.ts#L18

I can investigate more tonight...

jasonjurotich:
nope, just did:
clasp create (name)
clasp clone (to download the Code.js file)
then I opened VIM to add the one line of code. 
Then clasp run myFunction
... It did not work, so I opened it like normal, ran it to see if it was working, and it was.
Then went back and put clasp run myFunction 
and the error still remained. 

Everything else in clasp works just fine. I have not done any deployment. I just wanted to do the simplest thing possible to see if it would work. 

I wanted to do a YouTube video on this to promote it, but, well, haha.... hopefully this can be fixed so as to promote what you are doing. Thanks!!!!

jasonjurotich:
I forgot to add that I also pushed everything, before opening it up like normal to test it, and I also used sudo clasp run myFunction, but the error always remains (just did the process again).

grant:
@jasonjurotich Could you file a new bug with more details about the error your seeing with the setup instructions? 
Clasp run is pretty new so it may have errors, but it should work.

jasonjurotich:
With pleasure. Would that be under Issues here or is there somewhere else I should put it? (a link would be nice if it were not here). Thanks.

grant:
Please file an issue here in `clasp` since you had an issue with `clasp run`.
[Link](https://github.com/google/clasp/issues/new)

Issue #105, opened 2018-04-11
Make clasp smaller


grant:
## Expected Behavior

Clasp is <1000 lines of code.

## Actual Behavior

`index.ts` is ~1.1k SLOCs.

## Ways to shrink logic
- Inline `getFileType`, `getAPIFileType`, `manifestExists`
- Remove `commander` boilerplate.
- Make `getAPICredentials` async.
- Make online support required (unless specified).
- Make google API requests async
- Combine API call callback.
- Remove/shorten:
```ts
spinner.stop(true);
if (error) {
  logError(error);
}
```
- Extract dotfile logic in separate npm module.
- But keep to 1 file.

sqrrrl:
Just curious -- why impose a 1000 loc limit on clasp?

grant:
This issue is more of a suggestion. I think it would be good to extract some logic into npm modules.
I think ~1k is about the spot where a developer can fully understand what is happening within the code.
Also I've noticed `tsc` is getting slower the more lines we add.

grant:
I think we'll break apart the single file with #134.
But there are some good ideas of keeping `clasp` small here.

campionfellin:
This is being worked on in #133 and #134. Going to close this issue in favor of more specific issues like the others.

PR #106, opened 2018-04-12
Move basic things to utils file

campionfellin:
This commit factors out the project name, project manifest basename, dot, and dotfile constants to a `utils.ts` file.

A good start to working on re-factoring and cleaning up the code.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -45,30 +45,7 @@ const readline = require('readline');
 import { Server } from "http";
 const logging = require('@google-cloud/logging');
 const chalk = require('chalk');
-
-// Names / Paths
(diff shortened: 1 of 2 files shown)


grant:
Can we keep the current project in one file? Is one file a pain?

I think there are some other things we should do before breaking apart the project to multiple files.

Ideally, PRs should be backed by a GitHub Issue.

campionfellin:
Yeah one file is fine! I was looking at https://github.com/google/clasp/issues/105 just as an easy way to cut down on size.

grant:
Got it.

PR #107, opened 2018-04-12
Fix clasp logs

campionfellin:
To run this, clone the latest version of `clasp`. Make these few changes. Now say you `clasp clone` or `clasp create` a project. (Better to use `clasp clone` with one that already has been run, as sometimes StackDriver logs take some time to update). Then when you run `clasp logs` you'll get:

```
Please set your projectId in your .clasp.json file to your Google Cloud project ID.

  You can find your projectId by following the instructions in the README here:

  https://github.com/google/clasp#get-project-id
```

Basically do `clasp open` then hit Resources -> Cloud Platform Project and copy your `project-id-xxx`

Then put it in your `.clasp.json` like so:

```
  {
    "scriptId":"14Ht4FoesbNDhRbbTMI_IyM9uQ27EXIP_p2rK8xCOECg5s9XKpHp4fh3d",
    "projectId": "project-id-xxxxxxxxxxxxxxxxxxx"
  }
```

run `clasp logs` and you should see something like:

```
ERROR Sat Apr 07 2018 10:58:31 GMT-0700 (PDT) myFunction      oh no
INFO  Sat Apr 07 2018 10:58:31 GMT-0700 (PDT) myFunction      new
DEBUG Sat Apr 07 2018 10:58:31 GMT-0700 (PDT) myFunction      suppp
ERROR Sat Apr 07 2018 10:58:30 GMT-0700 (PDT) myFunction      oh no
INFO  Sat Apr 07 2018 10:58:30 GMT-0700 (PDT) myFunction      new
```

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1049,8 +1049,6 @@ commander
     open: boolean,
   }) => {
     await checkIfOnline();
-    console.log('IN DEVELOPMENT!');
-    process.exit(0);
     function printLogs([entries]:any[]) {
       for (let i = 0; i < 5; ++i) {
         const metadata = entries[i].metadata;
@@ -1082,7 +1080,7 @@ commander
       }
     }
 
-    getProjectSettings().then(({ projectId }: ProjectSettings) => {
+    getProjectSettings().then(({ scriptId, rootDir, projectId }: ProjectSettings) => {
       if (!projectId) {
         console.error(ERROR.NO_GCLOUD_PROJECT);
         process.exit(-1);
@@ -1095,10 +1093,8 @@ commander
       }
       const logger = new logging({
         projectId,
-        // auth: oauth2Client
       });
-      console.log(logger.getEntries());
-      // return logger.getEntries().then(printLogs);
+      return logger.getEntries().then(printLogs);
     });
   });
 


campionfellin:
Hey, no worries. When I echo `$GOOGLE_APPLICATION_CREDENTIALS` I don't get anything. Another thing to troubleshoot would be to open https://console.cloud.google.com/apis/dashboard?project=project-id-xxxxx and check to make sure that Apps Scripts API is enabled. Could be that I enabled it previously and forgot to put that in the README. Let me know if that works.

grant:
The Apps Script API is enabled.
But the `logger` doesn't have any credentials. How does it make auth'd requests?

https://github.com/google/clasp/blob/master/index.ts#L1095-L1097

Observe no `auth`.

## Docs

https://cloud.google.com/docs/authentication/getting-started

> // Instantiates a client. If you don't specify credentials when constructing
> // the client, the client library will look for credentials in the
> // environment.

For your machine, you must have keys somewhere that the npm module uses.

See env variables with `printenv`

campionfellin:
Hey, this is my last attempt here before I call it a night. Right before this line:
https://github.com/google/clasp/blob/3b56999b895e687406a8d5d8cd51d197d4792748/index.ts#L1098

Put `console.log(logger);`
It appears that at one point, I gave it allllll the scopes:
```
Logging {
  api: {},
  auth:
   Auth {
     authClientPromise: null,
     authClient: null,
     googleAuthClient: null,
     config:
      { libName: 'gccl',
        libVersion: '1.2.0',
        scopes: [Array],
        projectId: 'project-id-2758325992214167277' },
     credentials: null,
     environment: {},
     jwtClient: null,
     projectId: 'project-id-2758325992214167277' },
  options:
   { libName: 'gccl',
     libVersion: '1.2.0',
     scopes:
      [ 'https://www.googleapis.com/auth/cloud-platform',
        'https://www.googleapis.com/auth/cloud-platform.read-only',
        'https://www.googleapis.com/auth/logging.admin',
        'https://www.googleapis.com/auth/logging.read',
        'https://www.googleapis.com/auth/logging.write',
        'https://www.googleapis.com/auth/cloud-platform',
        'https://www.googleapis.com/auth/cloud-platform.read-only',
        'https://www.googleapis.com/auth/logging.admin',
        'https://www.googleapis.com/auth/logging.read',
        'https://www.googleapis.com/auth/logging.write',
        'https://www.googleapis.com/auth/cloud-platform',
        'https://www.googleapis.com/auth/cloud-platform.read-only',
        'https://www.googleapis.com/auth/logging.admin',
        'https://www.googleapis.com/auth/logging.read',
        'https://www.googleapis.com/auth/logging.write' ],
     projectId: 'project-id-2758325992214167277' },
  projectId: 'project-id-2758325992214167277' }
```

So I'm guessing this is why. To test this, go here https://github.com/google/clasp/blob/3b56999b895e687406a8d5d8cd51d197d4792748/index.ts#L320 and add a handful of those scopes, then `clasp logout`, `clasp login` to give access to all these. We probably want to narrow it down (especially as it looks like it repeats the same 5 here), but see if that's the issue.

Otherwise, maybe time to put some effort into using something like Docker so we're all working with the same environment?

grant:
Oh, we probably need the [OAuth scope](https://developers.google.com/identity/protocols/googlescopes#loggingv2):
`https://www.googleapis.com/auth/logging.read`
Not sure how it worked without that.

Still getting the Unhandled promise rejection 5 times. I'm not sure why it would log 5 times even after changing the number of logs. I'll have to investigate on another machine or something.

campionfellin:
Hey, yeah some weird things are going on here for sure. One thing I saw online about credentials is:
```
Providing credentials to your application
GCP client libraries use a strategy called Application Default Credentials (ADC) to find your application's credentials. When your code uses a client library, the strategy checks for your credentials in the following order:

1. First, ADC checks to see if the environment variable GOOGLE_APPLICATION_CREDENTIALS is set. If the variable is set, ADC uses the service account file that the variable points to.

2. If the environment variable isn't set, ADC uses the default service account that Compute Engine, Kubernetes Engine, App Engine, and Cloud Functions provide, for applications that run on those services.

3. If ADC can't use either of the above credentials, an error occurs.
```

Maybe since I don't have that environmental variable set, it goes to 2 and finds a service account saved somewhere in ~/

PR #111, opened 2018-04-17
Re-format .travis.yml

campionfellin:
Adds `node_modules` to cache, so it won't reinstall unless there are new ones to add.
Removes node versions < 6
Installs and builds before linting/testing
After testing runs coverage report

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: required
 os:
   - linux
   - osx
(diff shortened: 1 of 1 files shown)


campionfellin:
Looks like build failed for linux because build starts with `rm index.js`. Is there a reason we do this? Doesn't it just overwrite it when we run `tsc --project tsconfig.json` anyway?

PR #118, opened 2018-04-18
Add npm ci check

ajlai:
npm 5.8.0 has a new command [npm-ci](https://docs.npmjs.com/cli/ci) which functions like `npm install` but also checks that the package-lock.json is up to date with the package.json file.

This PR changes the travis config to use `npm ci`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,15 @@ node_js:
   - '8'
   - '7'
   - '6'
+# keep the npm cache around to speed up installs
 cache:
   directories:
-  - node_modules
+  - "$HOME/.npm"
+before_install:
+  - npm install -g npm@latest
+install:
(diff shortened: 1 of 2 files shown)


ajlai:
For an example of how a travis failure looks when package-lock.json hasn't been updated, see https://travis-ci.org/ajlai/clasp/jobs/367926019

grant:
I appreciate the PR, but I don't think this PR speeds up the build.
https://travis-ci.org/google/clasp/builds/367971284

I'd remove `osx` / `linux` if the goal is to speed up Travis.

ajlai:
My original goal wasn't to speed up the build (although that would have been nice). It's just to make sure package-lock.json gets updated and checked in when the package.json does.

PR #119, opened 2018-04-18
Creates tests for clasp clone and clasp pull

campionfellin:
When running clasp clone test make sure you put in a script ID.

### DIFF
--- a/tests/test.ts
+++ b/tests/test.ts
@@ -1,5 +1,6 @@
 import { describe, it } from 'mocha';
 import { expect } from 'chai';
+import * as fs from 'fs';
 const { spawnSync } = require('child_process');
 
 describe('Test help for each function', () => {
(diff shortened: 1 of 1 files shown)


tests/test.ts
```diff
@@ -43,6 +43,31 @@ describe.skip('Test clasp create function', () => {
   });
 });
 
+// When running this test, please put in a script ID
```
grant: What do you mean "put in a script ID"?
Can we first create a test script ID, then clone it?

tests/test.ts
```diff
@@ -43,6 +43,31 @@ describe.skip('Test clasp create function', () => {
...
+// then use `sudo npm run build` before running test.
```
grant: Don't you need to build before any test?

Shouldn't we have this in the file header comment if that is the case?

grant:
Not sure where the `-61.02%` came from.

tests/test.ts
campionfellin: Yeah, we can do one and hard code it for everyone, if you think that's better. Or use the one from the clasp create test.

tests/test.ts
campionfellin: Yeah good point

campionfellin:
Not sure on the coveralls issue either...

campionfellin:
So this is because some of the tests can't be run on Travis so they are skipped. If you run locally and un-skip every test, this is the output:

```
-------------|----------|----------|----------|----------|-------------------|
File         |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-------------|----------|----------|----------|----------|-------------------|
All files    |    36.14 |    24.05 |    33.66 |    36.89 |                   |
 clasp       |    32.48 |    24.05 |     29.9 |    32.89 |                   |
  index.js   |    32.48 |    24.05 |     29.9 |    32.89 |... 1285,1294,1298 |
 clasp/tests |      100 |      100 |      100 |      100 |                   |
  test.js    |      100 |      100 |      100 |      100 |                   |
-------------|----------|----------|----------|----------|-------------------|
```

Issue #124, opened 2018-04-21
Clasp logout doesn't logout when you use clasp login --ownkey


campionfellin:
## Expected Behavior

`clasp logout` should delete `.clasprc.json` in working directory (and in ~/)

## Actual Behavior

`clasp logout` only deletes from `~/`, see here: https://github.com/google/clasp/blob/3da3a3c69171aaa430d19852f6f2f15031f6f5e8/index.ts#L567

Probably just need to add a line after with `del(DOT.RC_LOCAL.ABSOLUTE_PATH, { force: true });`

## Steps to Reproduce the Problem

  1. `clasp login && cat ~/.clasprc.json && clasp logout && cat ~/.clasprc.json`
  1. `clasp login --ownkey && cat ./.clasprc.json && clasp logout && cat ./.clasprc.json`

## Specifications

  - Node version (`node -v`): v8.9.4
  - Version (`npm list -g | grep clasp`): dev build
  - OS (Mac/Linux/Windows): macOS

PR #125, opened 2018-04-21
Adds clasp logout test

campionfellin:
Note: this test will fail if you login using clasp login --ownkey

See issue: https://github.com/google/clasp/issues/124

This doesn't fix it, but is related.

### DIFF
--- a/tests/test.ts
+++ b/tests/test.ts
@@ -100,6 +100,20 @@ describe.skip('Test clasp open function', () => {
   });
 });
 
+// Fails when you logged in using --ownkey flag
+describe.skip('Test clasp logout function', () => {
+  it('should logout correctly', () => {
+    const result = spawnSync(
+      'clasp', ['logout'], { encoding: 'utf8' }
+    );
+    expect(result.status).to.equal(0);
+    const localDotExists = fs.existsSync('.clasprc.json');
+    expect(localDotExists).to.equal(false);
+    const dotExists = fs.existsSync('~/.clasprc.json');
+    expect(dotExists).to.equal(false);
+  });
+});
+
 /**
  * TODO: Test these commands and configs.
  *
@@ -107,7 +121,7 @@ describe.skip('Test clasp open function', () => {
  * [ ] clasp;
  * [ ] clasp login';
  * [ ] clasp login --no-localhost;
- * [ ] clasp logout;
+ * [x] clasp logout;
  * [x] clasp create "myTitle"
  * [x] clasp create <untitled>
  * [x] clasp list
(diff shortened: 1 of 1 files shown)


tests/test.ts
```diff
@@ -100,14 +100,28 @@ describe.skip('Test clasp open function', () => {
...
+describe.skip('Test clasp logout function', () => {
+  it('should logout correctly', () => {
+    const result = spawnSync(
+      'clasp', ['logout'], { encoding: 'utf8' }
+    );
+    expect(result.status).to.equal(0);
+    const localDotExists = fs.existsSync('.clasprc.json');
+    expect(localDotExists).to.equal(false);
+    const dotExists = fs.existsSync('~/.clasprc.json');
+    expect(dotExists).to.equal(false);
```
grant: Shouldn't
`expect(dotExists).to.equal(true);`
exist somewhere in this test?

campionfellin:
Also, this is the new test coverage:
```
-------------|----------|----------|----------|----------|-------------------|
File         |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-------------|----------|----------|----------|----------|-------------------|
All files    |    49.03 |    34.26 |    49.53 |     50.3 |                   |
 clasp       |    44.12 |    34.26 |    44.56 |    45.03 |                   |
  index.js   |    44.12 |    34.26 |    44.56 |    45.03 |... 1286,1295,1299 |
 clasp/tests |      100 |      100 |      100 |      100 |                   |
  test.js    |      100 |      100 |      100 |      100 |                   |
-------------|----------|----------|----------|----------|-------------------|
```

tests/test.ts
campionfellin: We should expect that `clasp logout` removes the `.clasprc.json` file. So if you want, I can put `expect(dotExists).to.equal(true);` before we run the `clasp logout` command? To check if it already exists. Though if you run `clasp logout` and you weren't logged in, nothing happens anyway.

campionfellin: Alternatively, maybe we want to create a command `clasp logout --ownkey`, in which it deletes the key in the working directory. Then we could say `expect(dotExists).to.equal(true);`

grant: I think this test would still pass if you removed the `result`, seems like a NoOp.

It doesn't really do anything, the rc files were never there in the first place, right?

This test is better than nothing, but I'd first create a rc file, then see if logging out deletes it.

campionfellin: Yes that is correct, it really doesn't do anything without the `clasp login` test. If you manually run `clasp login --ownkey` to create the rc file in the working directory and then run this test, it will fail. If you manually run `clasp login` to create the rc file in the home directory, and then run this test it will pass.

I'll just touch an rc file before the tests, one sec.

ajlai:
https://github.com/google/clasp/issues/126

PR #127, opened 2018-04-21
Fix language identification of .ts files

ajlai:
This is a nitpicky fix for Github misidentifying `index.ts` as a javascript file:
image.

I tested with linguist locally (https://github.com/github/linguist#application-usage) to confirm that this change fixes the issue.


grant:
Thanks!
The reason why the project was ID'd as Node was because of the shebang (`#!/usr/bin/env node`).

Issue #128, opened 2018-04-22
Clasp logs sometimes returns unhandled promise rejection


campionfellin:
## Expected Behavior

`clasp logs` --> it shows your logs

## Actual Behavior

Sometimes when running `clasp logs` you get an unhandled promise rejection that looks like:

```
(node:2477) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Retry total timeout exceeded before anyresponse was received
(node:2477) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

## Steps to Reproduce the Problem

  1. `clasp create`
  1.  Run your function that logs things, like `console.log('logging now...')`

## Specifications

  - Node version (`node -v`): v8.9.4
  - Version (`npm list -g | grep clasp`): dev version
  - OS (Mac/Linux/Windows): macOS

PR #129, opened 2018-04-22
Fixes unhandled rejection in clasp logs

campionfellin:
Does this by catching the error and logging an appropriate message for the user.

The error that it catches looks like:

```
{ Error: 8 RESOURCE_EXHAUSTED: Insufficient tokens for quota 'logging.googleapis.com/read_requests' and limit 'ReadRequestsPerMinutePerUser' of s
ervice 'logging.googleapis.com' for consumer 'project_number:xxxxxxxx'.
    at new createStatusError (/Users/campionfellin/Desktop/gcp-work/clasp/node_modules/grpc/src/client.js:64:15)
    at /Users/campionfellin/Desktop/gcp-work/clasp/node_modules/grpc/src/client.js:583:15
  code: 8,
  metadata:
   Metadata {
     _internal_repr:
      { 'google.rpc.help-bin': [Array],
```

My guess is that this error comes about because StackDriver logging takes a few minutes to get all set up. This PR catches that error and tells the user to try again.

I also noticed another severity level of `NOTICE` and so added that as well.

Closes https://github.com/google/clasp/issues/128

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -190,6 +190,7 @@ Forgot ${PROJECT_NAME} commands? Get help:\n  ${PROJECT_NAME} --help`,
   FS_FILE_WRITE: 'Could not write file.',
   LOGGED_IN: `You seem to already be logged in. Did you mean to 'logout'?`,
   LOGGED_OUT: `\nCommand failed. Please login. (${PROJECT_NAME} login)`,
+  LOGS_UNAVAILABLE: 'StackDriver logs are getting ready, try again soon.',
   OFFLINE: 'Error: Looks like you are offline.',
   ONE_DEPLOYMENT_CREATE: 'Currently just one deployment can be created at a time.',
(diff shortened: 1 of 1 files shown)


index.ts
```diff
@@ -1071,9 +1072,10 @@ commander
         let coloredSeverity = ({
           ERROR: chalk.red(severity),
           INFO: chalk.blue(severity),
-          DEBUG: chalk.yellow(severity)
+          DEBUG: chalk.yellow(severity),
+          NOTICE: chalk.magenta(severity)
```
grant: (We should add [`trailing-comma`](https://palantir.github.io/tslint/rules/trailing-comma/) to our `tslint` to prevent 2 line diffs.)

campionfellin: Noted, will put that in the next PR.

Issue #133, opened 2018-04-24
Move utility functions to src/utils.ts


campionfellin:
This will improve the readability and maintainability of `clasp`

* [ ] Move util objects and functions like DOT, DOTFILE, etc to `src/utils.ts`
* [ ] Integrate changes into `index.ts`
* [ ] Run all the tests (all are relevant)

grant:
SGTM.
I'd probably do this first before moving the commands and make sure the `tsc` command is correct.

campionfellin:
Yeah this issue is especially important to improving our test coverage. Once separated, it should be much easier to test methods like: https://github.com/google/clasp/blob/9ad0a4220b6b84b9a9eb6a511297e24461e2dae5/index.ts#L263

PR #135, opened 2018-04-25
Breaks out some of the consts from index.ts to utils.ts

campionfellin:
Starting work on https://github.com/google/clasp/issues/133

Passes lint and all tests.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -11,6 +11,7 @@
   "moduleResolution": "node",
   "files": [
     "index.ts",
-    "tests/test.ts"
+    "tests/test.ts",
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -11,6 +11,7 @@
   "moduleResolution": "node",
   "files": [
     "index.ts",
-    "tests/test.ts"
+    "tests/test.ts",
+    "src/utils.ts"
```
grant: We should take a `src/*.ts` glob next time.
http://www.typescriptlang.org/docs/handbook/tsconfig-json.html

PR #141, opened 2018-04-25
Adds tests for clasp deploy and clasp deployments

campionfellin:
Additionally, gives a better error message for when a deployment fails and fixes an unhandled promise error.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -116,7 +116,7 @@ const ERROR = {
 Forgot ${PROJECT_NAME} commands? Get help:\n  ${PROJECT_NAME} --help`,
   CONFLICTING_FILE_EXTENSION: (name: string) => `File names: ${name}.js/${name}.gs conflict. Only keep one.`,
   CREATE: 'Error creating script.',
-  DEPLOYMENT_COUNT: `Unable to deploy; Only one deployment can be created at a time`,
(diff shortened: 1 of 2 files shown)


index.ts
```diff
@@ -116,7 +116,7 @@ const ERROR = {
 Forgot ${PROJECT_NAME} commands? Get help:\n  ${PROJECT_NAME} --help`,
   CONFLICTING_FILE_EXTENSION: (name: string) => `File names: ${name}.js/${name}.gs conflict. Only keep one.`,
   CREATE: 'Error creating script.',
-  DEPLOYMENT_COUNT: `Unable to deploy; Only one deployment can be created at a time`,
+  DEPLOYMENT_COUNT: `Unable to deploy; Scripts may only have up to 20 versioned deployments at a time.`,
```
campionfellin: I could not get the previous error message, but I often got this. If you think they should be 2 separate messages, I can change that.

index.ts
```diff
@@ -821,12 +821,12 @@ commander
               manifestFileName: PROJECT_MANIFEST_BASENAME,
               description,
             },
-          }, {}, (err: any, { data }: any) => {
+          }, {}, (err: any, response: any) => {
```
campionfellin: So, when it returns an `err`, it actually returns `undefined` as the response, so when using `{ data }` you get an unhandled promise rejection error. This fixes that.

PR #143, opened 2018-04-25
Add lint to PR template

campionfellin:


grant:
Why is Coveralls running 6 times?

campionfellin:
Not sure, I think maybe it is just commenting 6 times? We could turn it off altogether.

PR #144, opened 2018-04-25
Allow clasp open to work for any scriptId

### LINKED ISSUE #100: clasp open [scriptId]
## Expected Behavior

Now that you can `clasp list`, you should be able to open a specific `scriptId`.
`clasp open [scriptId]`

## Actual Behavior

`clasp open` doesn't allow params.

campionfellin:
Adds optional parameter [scriptId] to clasp open command.

Fixes #100 

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ clasp deployments # List all deployment IDs
 clasp open [scriptId]
 ```
 
-Takes an optional argument of the scriptId that you'd like to open. Else, opens script from `.clasp.json`
+Opens the `clasp` project on script.google.com. Provide a `scriptId` to open a different script.
 
 ### List your App Scripts (In Development)
 
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ clasp
 - `clasp clone <scriptId>`
 - `clasp pull`
 - `clasp push`
-- `clasp open`
+- `clasp open [scriptId]`
 - `clasp deployments`
 - `clasp deploy [version] [description]`
 - `clasp redeploy <deploymentId> <version> <description>`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -107,9 +107,11 @@ clasp deployments # List all deployment IDs
 ### Open the Project on script.google.com
 
 ```
-clasp open
+clasp open [scriptId]
 ```
 
+Takes an optional argument of the scriptId that you'd like to open. Else, opens script from `.clasp.json`
```
grant: First describe what it does without options (90% of cases). Then describe options.
```
Opens the `clasp` project on script.google.com. Provide a `scriptId` to open a different script.
```

index.ts
```diff
@@ -628,19 +628,26 @@ commander
...
+      } else {
+        console.log(LOG.OPEN_PROJECT(scriptId));
+        open(getScriptURL(scriptId));
       }
-    });
+    };
+    if (scriptId) {
+      openScript(scriptId);
+    } else {
+      getProjectSettings().then(async ({ scriptId }: ProjectSettings) => {
```
grant: Let's write this:

```ts
getProjectSettings().then(openScript);
```

```ts
const openScript = (scriptId?) => {
  if (!scriptId) return;
  // if (scriptId.length < 30) { ... length check
  // } else { ... log and open
}
```

In the future (when `getProjectSettings` is async):

```ts
const openScript = (scriptId?) => { ... }
openScript(scriptId ? scriptId : (await getProjectSettings()).scriptId);
```

grant: #146 Created issue for async getProjectSettings

PR #145, opened 2018-04-25
Add test for clasp versions

campionfellin:

### DIFF
--- a/tests/test.ts
+++ b/tests/test.ts
@@ -142,6 +142,17 @@ describe.skip('Test clasp deploy function', () => {
   });
 });
 
+describe.skip('Test clasp versions function', () => {
+  it('should list versions correctly', () => {
+    const result = spawnSync(
+      'clasp', ['versions'], { encoding: 'utf8' },
+    );
+    expect(result.stdout).to.contain('~ ');
+    expect(result.stdout).to.contain(' Versions ~');
+    expect(result.status).to.equal(0);
+  });
+});
+
 // Fails when you logged in using --ownkey flag
 describe.skip('Test clasp logout function', () => {
   it('should logout correctly', () => {
@@ -179,7 +190,7 @@ describe.skip('Test clasp logout function', () => {
  * [ ] clasp deploy [version] [description]
  * [ ] clasp redeploy <deploymentId> <version> <description>
  * [ ] clasp version [description]
- * [ ] clasp versions
+ * [x] clasp versions
  *
  * # Configs
  * - .js and .gs files
--- a/tests/test.ts
+++ b/tests/test.ts
@@ -2,6 +2,7 @@ import { describe, it } from 'mocha';
 import { expect } from 'chai';
 import * as fs from 'fs';
 const { spawnSync } = require('child_process');
+import { getScriptURL, getFileType } from './../src/utils.js';
 
 describe('Test help for each function', () => {
   it('should output help for run command', () => {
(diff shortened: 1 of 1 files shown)


tests/test.ts
```diff
@@ -142,6 +143,37 @@ describe.skip('Test clasp deploy function', () => {
...
+    const url = getScriptURL('abcdefghijklmnopqrstuvwxyz');
+    expect(url).to.equal('https://script.google.com/d/abcdefghijklmnopqrstuvwxyz/edit');
+    const url2 = getScriptURL('1234567890abcdefg');
+    expect(url2).to.equal('https://script.google.com/d/1234567890abcdefg/edit');
+  });
+});
+
+describe.skip('Test getFileType function from utils', () => {
+  it('should return the lowercase file type correctly', () => {
+    const type = getFileType('SERVER_JS');
```
campionfellin: This function only appears in `index.ts` once. We could probably inline it, as all it does is a `toLowerCase` unless the filetype is `SERVER_JS`, where it then just converts it to `js`

tests/test.ts
```diff
@@ -142,6 +143,37 @@ describe.skip('Test clasp deploy function', () => {
   });
 });
 
+describe.skip('Test clasp versions function', () => {
+  it('should list versions correctly', () => {
+    const result = spawnSync(
+      'clasp', ['versions'], { encoding: 'utf8' },
+    );
+    expect(result.stdout).to.contain('~ ');
+    expect(result.stdout).to.contain(' Versions ~');
```
grant: This doesn't seem like a very strong test. Can we create a version, then ensure the version we created is the same as the version listed?

tests/test.ts
```diff
@@ -142,6 +143,37 @@ describe.skip('Test clasp deploy function', () => {
...
+    expect(result.status).to.equal(0);
+  });
+});
+
+describe.skip('Test getScriptURL function from utils', () => {
+  it('should return the scriptURL correctly', () => {
+    const url = getScriptURL('abcdefghijklmnopqrstuvwxyz');
+    expect(url).to.equal('https://script.google.com/d/abcdefghijklmnopqrstuvwxyz/edit');
+    const url2 = getScriptURL('1234567890abcdefg');
+    expect(url2).to.equal('https://script.google.com/d/1234567890abcdefg/edit');
```
grant: This URL doesn't test anything different than the previous URL.
Omit it.

tests/test.ts
```diff
@@ -142,6 +143,37 @@ describe.skip('Test clasp deploy function', () => {
...
+  });
+});
+
+describe.skip('Test getFileType function from utils', () => {
+  it('should return the lowercase file type correctly', () => {
+    const type = getFileType('SERVER_JS');
+    expect(type).to.equal('js');
+    const type2 = getFileType('GS');
```
grant: Inline all `const`s in this function. They're only used once.

tests/test.ts
grant: This function is fine standalone. It's good to test it for coverage.

tests/test.ts
```diff
@@ -142,6 +143,42 @@ describe.skip('Test clasp deploy function', () => {
...
 
+describe.skip('Test clasp version and versions function', () => {
+  let versionNumber = '';
+  it('should create new version correctly', () => {
+    const result = spawnSync(
+      'clasp', ['version'], { encoding: 'utf8' },
+    );
+    expect(result.stdout).to.contain('Created version ');
+    expect(result.status).to.equal(0);
+    versionNumber = result.stdout.substring(result.stdout.lastIndexOf(" "), result.stdout.length-2);
```
campionfellin: Gets version number of version created

tests/test.ts
```diff
@@ -142,6 +143,42 @@ describe.skip('Test clasp deploy function', () => {
...
+    it('should list versions correctly', () => {
+      const result = spawnSync(
+        'clasp', ['versions'], { encoding: 'utf8' },
+      );
+      expect(result.stdout).to.contain('~ ');
+      expect(result.stdout).to.contain(' Versions ~');
+      expect(result.stdout).to.contain(versionNumber + ' - ');
```
campionfellin: And checks if that version is listed.

tests/test.ts
grant: tslint should've checked for spaces aside the `-` operator.

tests/test.ts
```diff
@@ -142,6 +143,42 @@ describe.skip('Test clasp deploy function', () => {
...
+      'clasp', ['version'], { encoding: 'utf8' },
+    );
+    expect(result.stdout).to.contain('Created version ');
+    expect(result.status).to.equal(0);
+    versionNumber = result.stdout.substring(result.stdout.lastIndexOf(" "), result.stdout.length-2);
+    it('should list versions correctly', () => {
+      const result = spawnSync(
+        'clasp', ['versions'], { encoding: 'utf8' },
+      );
+      expect(result.stdout).to.contain('~ ');
```
grant: Ideally changing the `~` character (i.e. `–`) wouldn't break this test.
I suppose it's not that big of a deal if we change it later.

tests/test.ts
grant: Also, use `'` not `"`.

Issue #146, opened 2018-04-26
Make getProjectSettings async


grant:
`getProjectSettings` should be async to reduce nesting.

## Expected Behavior

```ts
let { scriptId, rootDir } = await getProjectSettings();
// ...
```

## Actual Behavior

```ts
getProjectSettings().then(({ scriptId, rootDir }: ProjectSettings) => {
  // ...
});
```

PR #152, opened 2018-04-27
Make getProjectSettings async

campionfellin:
Just for the functions we have tests for so far...

Also, put in a quick exit out of `clasp run` just in case anyone tries it.

Works on #146 but doesn't quite finish it...

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -414,23 +414,26 @@ commander
     } else {
       getAPICredentials(async () => {
         spinner.setSpinnerTitle(LOG.CREATE_PROJECT_START(title)).start();
-        getProjectSettings(true).then((settings: ProjectSettings) => {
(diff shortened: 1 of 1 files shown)


index.ts
```diff
@@ -414,23 +414,26 @@ commander
...
-            saveProjectId(scriptId);
-            if (!manifestExists()) {
-              fetchProject(scriptId); // fetches appsscript.json, o.w. `push` breaks
-            }
-          }).catch((error: object) => {
-            spinner.stop(true);
-            logError(error, ERROR.CREATE);
-          });
+        } catch (err) { // no scriptId (because project doesn't exist)
+          //console.log(err);
```
grant: Looks like a missing err handler.

PR #156, opened 2018-04-30
Fix build errors

campionfellin:
Now, there are less errors when running `sudo npm run build`.

The one that still persists is:

`Property 'redirectUri' is private and only accessible within class 'OAuth2Client'.`
 
Which will likely remain unless we change the way we make the `oauth2Client`.

Fixes #<issue_number_goes_here> (it's a good idea to open an issue first for discussion)

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -24,6 +24,7 @@ import 'connect';
 import * as del from 'del';
 import * as fs from 'fs';
 import { google } from 'googleapis';
+import { Drive } from 'googleapis/build/src/apis/drive/v3';
 import * as mkdirp from 'mkdirp';
 const open = require('open');
 const path = require('path');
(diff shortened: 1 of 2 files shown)


grant:
- What is the build error? What does this PR fix?
- Why does using the `[]` vs `.` syntax difference? I generally don't want to approve this change as this shouldn't matter or could be prevented.
- Why the es2017 change in this PR? I don't think we need all these libs.

campionfellin:
Hey, so when I compile with `sudo npm run build` I get all sorts of stuff like this:

screen shot 2018-04-30 at 2 58 36 pm

Which doesn't really affect anything in the end-product, but with these changes the output of `sudo npm run build` is:


screen shot 2018-04-30 at 3 00 33 pm

campionfellin:
The `[]` vs `.` syntax fixes the "Property files does not exist on type..." error.
Adding `es2017` gives us the `padEnd` function for strings, fixing "Property padEnd does not exist..."
And the big error with `BodyResponseCallBack` is fixed by removing `res: Function`

ajlai:
I was able to resolve the `.files` warning by adding an import
```typescript
import { Drive } from "googleapis/build/src/apis/drive/v3";
```
then adding the type information `as Drive` to `google.drive` usage.

Example:
```typescript
const drive = google.drive({version: 'v3', auth: oauth2Client}) as Drive;
```

Going that route instead of using `[]` allows VSCode to use the proper `Drive` type for referencing the `.files` method.

grant:
If @ajlai's resolutions are correct, that would be a huge improvement to this PR.

campionfellin:
@grant should I just edit this PR, or @ajlai do you want to create your own PR?

grant:
@campionfellin Given this PR is "Fix build errors", it seems appropriate that you can do that in this PR.

PR #158, opened 2018-05-02
Add more tests

campionfellin:
A few notes:

Looks like we should make saveProjectId  return a promise so we can use `await` syntax.

### DIFF
--- a/tests/test.ts
+++ b/tests/test.ts
@@ -2,7 +2,9 @@ import { describe, it } from 'mocha';
 import { expect } from 'chai';
 import * as fs from 'fs';
 const { spawnSync } = require('child_process');
-import { getScriptURL, getFileType } from './../src/utils.js';
+import { getScriptURL, getFileType, getAPIFileType,
+         saveProjectId } from './../src/utils.js';
+const path = require('path');
 
 describe('Test help for each function', () => {
   it('should output help for run command', () => {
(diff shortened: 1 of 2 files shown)


tests/test.ts
```diff
@@ -163,21 +165,41 @@ describe.skip('Test clasp version and versions function', () => {
...
+    expect(getAPIFileType('file.txt')).to.equal('TXT');
+  });
+});
+
+// NOTE: we should make saveProjectId async because dotf.write is async
+describe('Test saveProjectId function from utils', () => {
+  it('should save the scriptId correctly', () => {
+    spawnSync('rm', ['.clasp.json']);
+    saveProjectId('12345');
+    setTimeout(() => {
```
campionfellin: This is here because otherwise, it won't have saved the projectId before running the test.

tests/test.ts
```diff
@@ -163,21 +165,41 @@ describe.skip('Test clasp version and versions function', () => {
...
+    expect(getAPIFileType('file.JS')).to.equal('SERVER_JS');
+    expect(getAPIFileType('file.txt')).to.equal('TXT');
+  });
+});
+
+// NOTE: we should make saveProjectId async because dotf.write is async
+describe('Test saveProjectId function from utils', () => {
+  it('should save the scriptId correctly', () => {
+    spawnSync('rm', ['.clasp.json']);
+    saveProjectId('12345');
```
campionfellin: Should this be renamed to `saveScriptId`?

tests/test.ts
campionfellin: Yeah I was just putting that there so the tests didn't fail. I'll go ahead and make another commit to this branch with an update of `saveProjectId`

tests/test.ts
campionfellin: I'll change it from `saveProjectId` to `saveScriptId`. In my opinion, we should make a function at sometime that would support `clasp logs` by allowing users to add their GCP project ID to the `.clasp.json` file without having to manually edit it.

tests/test.ts
```diff
@@ -163,21 +165,41 @@ describe.skip('Test clasp version and versions function', () => {
...
   it('should return the lowercase file type correctly', () => {
     expect(getFileType('SERVER_JS')).to.equal('js');
     expect(getFileType('GS')).to.equal('gs');
     expect(getFileType('JS')).to.equal('js');
   });
 });
 
+describe('Test getAPIFileType function from utils', () => {
+  it('should return the uppercase file type correctly', () => {
+    expect(getAPIFileType('file.GS')).to.equal('SERVER_JS');
```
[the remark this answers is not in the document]
campionfellin: 👍

tests/test.ts
```diff
@@ -163,21 +165,41 @@ describe.skip('Test clasp version and versions function', () => {
...
+    expect(getAPIFileType('file.JS')).to.equal('SERVER_JS');
+    expect(getAPIFileType('file.txt')).to.equal('TXT');
+  });
+});
+
+// NOTE: we should make saveProjectId async because dotf.write is async
```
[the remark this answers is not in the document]
campionfellin: 👍

campionfellin:
All comments are covered except converting `saveProjectId` to `saveScriptId` because that needs to be changed in many places. Should be a separate PR.

Issue #131, opened 2018-04-24
Fix clasp login --ownkey and related issues


campionfellin:
## Expected Behavior

When you use `clasp login --ownkey`, all further `clasp` commands should use the `.clasprc.json` generated in your working directory. 

## Actual Behavior

Commands, like `clasp list`, `clasp login`, etc don't "know" about the `.clasprc.json` in the workdir. This is likely stemming from `getAPICredentials()`. We probably shouldn't pass in a boolean `isLocal`, but rather start the dotfile search in the current directory.

## Steps to Reproduce the Problem

  1. `clasp login --ownkey`
  1. `clasp list`
```
Could not read API credentials. Error:
{ Error: ENOENT: no such file or directory, open '/Users/campionfellin/.clasprc.json'
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/campionfellin/.clasprc.json' }
```

## Specifications

  - Node version (`node -v`): v8.9.4
  - Version (`npm list -g | grep clasp`): dev build
  - OS (Mac/Linux/Windows): macOS

PR #165, opened 2018-05-14
Fix clasp login

campionfellin:
Now it checks for local rc file before logging in.

Previously, if you ran `clasp login --ownkey` followed by `clasp login` you would have 2 `.clasprc.json` files, on in your homedir and one in your working directory. Now, it checks for a local rc file first.

Works on #131 (note, that we may have to implement some sort of check before every command so they do not always default to the rc file in your homedir.)

- [ ] `npm run test` succeeds. (no tests for login yet)

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -206,8 +206,12 @@ commander
     DOTFILE.RC.read().then((rc: ClaspSettings) => {
       console.warn(ERROR.LOGGED_IN);
     }).catch(async (err: string) => {
-      await checkIfOnline();
-      authorize(options.localhost, options.ownkey);
+      DOTFILE.RC_LOCAL.read().then((rc: ClaspSettings) => {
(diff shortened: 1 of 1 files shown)


grant:
Is this ready?

campionfellin:
I would say this part is. This prevents you from logging in using simply `clasp login` when you already have a `.clasprc.json` in your working directory.

The next step (whether this is a separate PR or on this PR) is to change the behavior of `getProjectSettings`
https://github.com/google/clasp/blob/0080f67565ae0af18baa027ffbfd29d794ad8d50/src/utils.ts#L154

It should check the current directory first for `.clasprc.json` and if not found then check the homedir.

Once this is done, commands like `clasp list`, etc should work after using `clasp login --ownkey`.
Do you think that should be with this or separate?

PR #172, opened 2018-05-16
Fix clasp run

campionfellin:
Steps to make work:

`clasp login`
`clasp create newScript`
`clasp open` (put `return "Hello, world!";` in `myFunction()`)

Click *Resources* -> *Cloud Platform Project* -> *Project* (should open up GCP)
Go to API Library and enable Apps Script API.
Create an OAuth client ID of type "Other"
Copy those credentials into here: https://github.com/google/clasp/blob/24c7d1309cde125b7a9db2a7d41f32bab4f5d7d1/src/auth.ts#L12
`sudo npm run build` to rebuild clasp with those credentials.
`clasp logout`

Go back to project in browser. Click *Publish* -> *Deploy as API executable...* then choose version and deploy.

`clasp login` to login with those credentials.
`clasp run myFunction`

Note: if this line https://github.com/google/clasp/blob/24c7d1309cde125b7a9db2a7d41f32bab4f5d7d1/index.ts#L822

is `false`, you should get a response like:
```
{ done: true,
  response:
   { '@type': 'type.googleapis.com/google.apps.script.v1.ExecutionResponse',
     result: 'HELLO' } }
```
else if it is `true` you'll get output that includes:
```
  code: 404,
  errors:
   [ { message: 'Requested entity was not found.',
       domain: 'global',
       reason: 'notFound' } ] }
```

- [ ] `npm run test` succeeds. (no tests yet)

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -808,20 +808,14 @@ commander
   .command('run <functionName>')
   .description('Run a function in your Apps Scripts project')
   .action((functionName) => {
-    console.log('IN DEVELOPMENT');
-    process.exit(0);
-    console.log('start run');
     getAPICredentials(async () => {
-      console.log('got creds');
       await checkIfOnline();
-      console.log('online');
       getProjectSettings().then(({ scriptId }: ProjectSettings) => {
         const params = {
           scriptId,
           function: functionName,
-          devMode: true,
+          devMode: false,
         };
-        console.log('about to run');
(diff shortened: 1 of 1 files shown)


campionfellin:
@grant I would say this is ready to merge except for the part where the user has to put in the credentials themselves. Is there a better way around this for now?

PR #177, opened 2018-05-21
Fix oauth2client redirectUri usage

ajlai:
This fixes the oauth2client usage warnings (
```
src/auth.ts(63,22): error TS2341: Property 'redirectUri' is private and only accessible within class 'OAuth2Client'.
src/auth.ts(80,18): error TS2341: Property 'redirectUri' is private and only accessible within class 'OAuth2Client'.
```
) by refactoring authorize functions and moves authorize into the auth module.

I've tested this change with the flows of clasp login / clasp login --no-localhost / clasp login --ownkey

With this, we'll be down to 0 build warnings! 🎉

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -1,19 +1,29 @@
 import { OAuth2Client } from 'google-auth-library';
-import { ClaspSettings, DOTFILE } from './utils.js';
+import { ClaspSettings, DOTFILE, ERROR } from './utils';
 import * as http from 'http';
 import * as url from 'url';
-const open = require('open');
-const readline = require('readline');
-import { LOG } from './commands.js';
+import open = require('open');
+import readline = require('readline');
+import { LOG } from './commands';
 
 // API settings
 // @see https://developers.google.com/oauthplayground/
-export const REDIRECT_URI_OOB = 'urn:ietf:wg:oauth:2.0:oob';
-export const oauth2Client = new OAuth2Client({
+const REDIRECT_URI_OOB = 'urn:ietf:wg:oauth:2.0:oob';
+const oauth2ClientAuthUrlOpts = {
+  access_type: 'offline',
+  scope: [
+    'https://www.googleapis.com/auth/script.deployments',
+    'https://www.googleapis.com/auth/script.projects',
+    'https://www.googleapis.com/auth/drive.metadata.readonly',
+    'https://www.googleapis.com/auth/script.webapp.deploy',
+  ],
+};
+const oauth2ClientSettings = {
   clientId: '1072944905499-vm2v2i5dvn0a0d2o4ca36i1vge8cvbn0.apps.googleusercontent.com',
   clientSecret: 'v6V3fKV_zWU7iw1DrpO1rknX',
   redirectUri: 'http://localhost',
-});
+};
+export const oauth2Client = new OAuth2Client(oauth2ClientSettings);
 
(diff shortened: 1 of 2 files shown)


src/auth.ts
```diff
@@ -1,19 +1,29 @@
 import { OAuth2Client } from 'google-auth-library';
-import { ClaspSettings, DOTFILE } from './utils.js';
+import { ClaspSettings, DOTFILE, ERROR } from './utils';
 import * as http from 'http';
 import * as url from 'url';
-const open = require('open');
-const readline = require('readline');
-import { LOG } from './commands.js';
+import open = require('open');
```
grant: Thanks for the cleanup.

src/auth.ts
```diff
@@ -38,61 +48,73 @@ export function getAPICredentials(cb: (rc: ClaspSettings | void) => void) {
...
-        open(authUrl);
-      });
+async function authorizeWithLocalhost() {
+  // Wait until the server is listening, otherwise we don't have
+  // the server port needed to set up the Oauth2Client.
+  const server = await new Promise<http.Server>((resolve, _) => {
+    const s = http.createServer();
+    s.listen(0, () => resolve(s));
+  });
+  const client = new OAuth2Client({...oauth2ClientSettings,
```
grant:
```ts
const client = new OAuth2Client({
  ...oauth2ClientSettings,
  redirectUri: `http://localhost:${server.address().port}`
});
```

src/auth.ts
```diff
@@ -38,61 +48,73 @@ export function getAPICredentials(cb: (rc: ClaspSettings | void) => void) {
...
-      });
-
-      server.listen(0, () => {
-        oauth2Client.redirectUri = `http://localhost:${server.address().port}`;
-        const authUrl = oauth2Client.generateAuthUrl(opts);
-        console.log(LOG.AUTHORIZE(authUrl));
-        open(authUrl);
-      });
+async function authorizeWithLocalhost() {
+  // Wait until the server is listening, otherwise we don't have
```
grant: Good comment.

ajlai:
Thanks for the speedy review @grant! I added the fixed error message to the PR description and pushed a change for const client comment.

I realized that auth could be further cleaned up by combining `getAPICredentials` and `oauth2Client`. We currently use them like:
```ts
const script = google.script({
  version: 'v1',
  auth: oauth2Client,
});
...
async function foo() {
  getAPICredentials(async () => {
    script.project.versions.list(...)
    ...
  }
}
```
however it's not clear that script/drive won't work correctly unless you've called getAPICredentials first. This could be clearer if script/drive usage was changed to:
```ts
// auth.ts
export function getAuthenticatedClient() : Promise<Oauth2Client> { ... }

// index.ts
async function getScript() {
  return google.script({
    version: 'v1',
    auth: await getAuthenticatedClient(),
  });
}
...
async function foo() {
    const script = await getScript();
    script.project.versions.list(...);
}
```
Thoughts on this?

grant:
```ts
import {getScript} from 'src/auth';
const script = await getScript();
```
Sounds good.

- We eventually want to move the commands from `index.ts` to `commands.ts` (1 PR).
- Here, `getScript` can be put in `auth.ts` too since it gets an auth'd `google.script` client. (1 PR).

Please create a GitHub issue with the plan, but that idea would be a good improvement!

Issue #181, opened 2018-05-23
Make clasp compile without errors with strict flag


ajlai:
## Expected Behavior
Clasp compiles without errors using --strict flag. The strict settings would help catch future bugs and make for a cleaner codebase.

## Actual Behavior
Compiling under --strict shows these errors:
```
[anthony@ambp] ~/dev/clasp master 2018-05-22 22:21:33 > npm run build

> @google/clasp@1.3.0 build /Users/anthony/dev/clasp
> tsc --project tsconfig.json; npm i -g;

index.ts(216,17): error TS7006: Parameter 'answers' implicitly has an 'any' type.
index.ts(218,17): error TS7006: Parameter 'err' implicitly has an 'any' type.
index.ts(313,17): error TS7034: Variable 'fileIds' implicitly has type 'any[]' in some locations where its type cannot be determined.
index.ts(316,15): error TS7005: Variable 'fileIds' implicitly has an 'any[]' type.
index.ts(322,25): error TS7005: Variable 'fileIds' implicitly has an 'any[]' type.
index.ts(323,23): error TS7006: Parameter 'answers' implicitly has an 'any' type.
index.ts(328,23): error TS7006: Parameter 'err' implicitly has an 'any' type.
index.ts(730,25): error TS7017: Element implicitly has an 'any' type because type '{ textPayload: any; jsonPayload: any; protoPayload: any; }' has no index signature.
index.ts(736,31): error TS7017: Element implicitly has an 'any' type because type '{ ERROR: any; INFO: any; DEBUG: any; NOTICE: any; }' has no index signature.
index.ts(761,55): error TS7006: Parameter 'err' implicitly has an 'any' type.
index.ts(778,12): error TS7006: Parameter 'functionName' implicitly has an 'any' type.
/Users/anthony/.npm-global/bin/clasp -> /Users/anthony/.npm-global/lib/node_modules/@google/clasp/index.js
+ @google/clasp@1.3.0
updated 1 package in 4.216s
[anthony@ambp] ~/dev/clasp master 2018-05-22 22:39:49 >
```

## Steps to Reproduce the Problem

  1.
  1.
  1.

## Specifications

  - Node version (`node -v`)
  - Version (`npm list -g | grep clasp`):
  - OS (Mac/Linux/Windows)

campionfellin:
Hey @ajlai , this is a great issue! Since we don't compile using --strict by default, I've put it in our `tsconfig` file here: https://github.com/google/clasp/pull/183
[...]

PR #183, opened 2018-05-24
Update tsconfig to include strict as a compiler option

campionfellin:
This will enable strict compiling by default, so that we can see errors like those brought up in #181 .

Related to #181 


grant:
Good PR comment, thanks!

Issue #134, opened 2018-04-24
Move commands to src/commands.ts


campionfellin:
This will improve the readability and maintainability of `clasp`

* [ ] Move commands one at a time to `src/commands.ts`
* [ ] `clasp login`
* [ ] `clasp logout`
* [ ] `clasp create`
* [ ] `clasp clone`
* [ ] `clasp pull`
* [ ] `clasp push`
* [ ] `clasp open`
* [ ] `clasp deployments`
* [ ] `clasp deploy`
* [ ] `clasp redeploy`
* [ ] `clasp version`
* [ ] `clasp versions`
* [ ] `clasp list`
* [ ] `clasp logs`
* [ ] Integrate changes into `index.ts`
* [ ] Run all the tests (all are relevant)

grant:
It would be cool if we reorganized the commands in a way where we could:
- Easily test commands
- Generate the command documentation
  - Since it's easy to miss optional args and new features

campionfellin:
I agree... I think that an even better organization would be something like:

`src/commands/login.ts`
`src/commands/logout.ts`
`src/commands/list.ts`

etc...

Thus each command could be easily isolated and tested.

campionfellin:
Yeah I like this idea, though a couple questions/comments:

1. Why would we be duplicating `LOG` and `ERROR`? Couldn't we just import them wherever we need?
2. I really like the idea of separating the command logic from the actual CLI library that we use, it'll be much easier to maintain and if we decide to switch to something like Vorpal that will make things easier.
3. This seems like quite a bit of work, should this be put into a major version? It might be more important to finish clasp 2.0 first, especially commands like `clasp run` before we work on this.

grant:
1. Assuming `commands.ts` uses all ~50 `LOG`/`ERROR`s, if we moved them to a separate file outside `commands`, we'd have to import/export them.
1. Separating SGTM. Maybe we can even `module.export` them sometime.
1. I think the next public version of clasp will be 2.0. I think the last main thing for the `2.0` list is `--watch`.

campionfellin:
1. Gotcha. I think it's also worth taking a look at reducing how many different `LOG`/`ERROR`s we have.
3. I think that much of this re-factoring should be pushed off to clasp 3.0 then (or maybe 2.1/2.2). In addition to `--watch`, we need to fix `clasp run`, as it doesn't seem to work at the moment. Probably related to finding a general solution to https://github.com/google/clasp/issues/131 / `getAPICredentials()`

[...]

PR #184, opened 2018-05-25
Move login function to commands.ts file

campionfellin:
This PR does a couple of things:

* Moves `login` functionality to `commands.ts` file. 
* Also adds `process.exit(0)` after `clasp login` so that it will end after logging in (otherwise after you login you need `CTRL+C` after logging in).

Works towards a model for #134 

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -164,12 +164,7 @@ commander
   .description('Log in to script.google.com')
   .option('--no-localhost', 'Do not run a local server, manually enter code instead')
   .option('--ownkey', 'Save .clasprc.json file to current working directory')
-  .action((options: {
-    localhost: boolean;
-    ownkey: boolean;
-  }) => {
-    login(options);
-  });
+  .action(login);
 
 /**
  * Logs out the user by deleteing client credentials.
--- a/index.ts
+++ b/index.ts
@@ -35,12 +35,12 @@ const logging = require('@google-cloud/logging');
 const chalk = require('chalk');
 const { prompt } = require('inquirer');
 import * as pluralize from 'pluralize';
-import { DOT, PROJECT_NAME, PROJECT_MANIFEST_BASENAME, ClaspSettings,
+import { DOT, PROJECT_NAME, PROJECT_MANIFEST_BASENAME,
     ProjectSettings, DOTFILE, spinner, logError, ERROR, getScriptURL,
     getProjectSettings, getFileType, getAPIFileType, checkIfOnline,
     saveProjectId, manifestExists } from './src/utils.js';
-import { oauth2Client, getAPICredentials, authorize } from './src/auth';
-import { LOG } from './src/commands.js';
+import { oauth2Client, getAPICredentials } from './src/auth';
+import { LOG, login } from './src/commands.js';
 // An Apps Script API File
 interface AppsScriptFile {
   name: string;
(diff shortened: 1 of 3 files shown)


index.ts
```diff
@@ -168,17 +168,7 @@ commander
...
-      console.warn(ERROR.LOGGED_IN);
-    }).catch(async (err: string) => {
-      DOTFILE.RC_LOCAL.read().then((rc: ClaspSettings) => {
-        console.warn(ERROR.LOGGED_IN);
-      }).catch(async (err: string) => {
-        await checkIfOnline();
-        authorize(options.localhost, options.ownkey);
-      });
-    });
+    login(options);
```
grant: If you look at this, we're calling a function with `options` that just calls a function with `options`.

I bet you can just do this:
```ts
commander
  .command('login')
  .description('Log in to script.google.com')
  .option('--no-localhost', 'Do not run a local server, manually enter code instead')
  .option('--ownkey', 'Save .clasprc.json file to current working directory')
  .action(login);
```

src/auth.ts
```diff
@@ -115,6 +115,7 @@ export async function authorize(useLocalhost: boolean, writeToOwnKey: boolean) {
     const token = await (useLocalhost ? authorizeWithLocalhost() : authorizeWithoutLocalhost());
     await (writeToOwnKey ? DOTFILE.RC_LOCAL.write(token) : DOTFILE.RC.write(token));
     console.log(LOG.AUTH_SUCCESSFUL);
+    process.exit(0);
```
grant: A comment as to why this is needed wouldn't hurt.

src/commands.ts
```diff
@@ -37,3 +39,16 @@ export const LOG = {
         (description || '(no description)'),
     VERSION_NUM: (numVersions: number) => `~ ${numVersions} ${pluralize('Version', numVersions)} ~`,
   };
+
+export function login(options) {
+    DOTFILE.RC.read().then((rc: ClaspSettings) => {
```
grant: Format indentation.

src/commands.ts
```diff
@@ -37,3 +39,16 @@ export const LOG = {
         (description || '(no description)'),
     VERSION_NUM: (numVersions: number) => `~ ${numVersions} ${pluralize('Version', numVersions)} ~`,
   };
+
+export function login(options) {
```
grant: Don't remove the param types.
```ts
export function login(options:{ localhost: boolean, ... }) {
```

src/commands.ts
```diff
@@ -37,3 +39,16 @@ export const LOG = {
         (description || '(no description)'),
     VERSION_NUM: (numVersions: number) => `~ ${numVersions} ${pluralize('Version', numVersions)} ~`,
   };
+
```
grant: Add a JSDoc comment.

campionfellin:
Will address those changes in a second here. Can you investigate why the travis build is failing? It seems like it only fails on Node (without specific version, maybe node 10?)

campionfellin:
This is why we need the `process.exit(0);`


clasp-login-prob

grant:
Re: The `process.exit`, yeah I know why it's added, but add a comment in the code. Also, is this in prod and should be patched ASAP?

Re: Travis failure
Looks like `grpc` uses `node-pre-gyp` which fails for some reason on the latest version of Node:

```
node-pre-gyp ERR! Tried to download(403): https://storage.googleapis.com/grpc-precompiled-binaries/node/grpc/v1.10.1/node-v64-linux-x64-glibc.tar.gz 
```

I don't know why this is though. Maybe try changing the grpc version or `package-lock.json`?

campionfellin:
Yes, looks like it was introduced here: https://github.com/google/clasp/pull/177/files

Scroll to bottom, line 117 of `src/auth.ts` (https://github.com/google/clasp/blob/6dc109bdd532efd761e2e6ae5a415041dcc39cd4/src/auth.ts#L117).

I'll finish this PR in a second here and we can merge it in.

src/commands.ts
campionfellin: 👍

src/auth.ts
campionfellin: 👍

ajlai:
Yikes, looks like hanging is from http_server.close() taking longer to gracefuly shut down. It looks like it takes ~4-5 seconds for me after authorizing for the clasp login command to close.

grant:
Had to restart the Travis build to get `grpc` to work.
For some reason it was still doing this `grpc@1.10.1 install /home/travis/build/google/clasp/node_modules/grpc`.

PR #191, opened 2018-05-26
Fix clasp clone with no scriptId given

### LINKED ISSUE #187: Issue with `clasp clone`
## Expected Behavior

It shows the list of scripts you can clone and you pick one and it clones it.

## Actual Behavior

Rather than trying to clone the `scriptId` it tries to clone `scriptName - (scriptId)`.

This is likely an error here: https://github.com/google/clasp/blob/02f8f278eda026b46cbfa23e8d4e51a2e20ee45b/index.ts#L315

That was introduced in https://github.com/google/clasp/pull/170

  1. `clasp clone`
  1.  Choose script
  1. It will spin for a while, with unhandled promise rejection
  1. If you `ctrl+c` and then `cat .clasp.json`, you'll see it saved it as `{"scriptId":"sampleProject        - (1osCvHwwtHXSf9BGxZkvG8dRlshwNlhLILqAtIbc_k0No0O9Yijrp3-NQ)"}` where it really should just be `{"scriptId":"1osCvHwwtHXSf9BGxZkvG8dRlshwNlhLILqAtIbc_k0No0O9Yijrp3-NQ"}`

## Specifications

  - Node version (`node -v`)
  - Version (`npm list -g | grep clasp`):
  - OS (Mac/Linux/Windows)

campionfellin:
Not sure that I like the hack-y way of getting the scriptId, but assuming that the scriptId itself shouldn't contain any `(` then I think it is valid, using `lastIndexOf("(")` for the start of the scriptId.

Fixes #187 

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -306,6 +306,8 @@ commander
               message : 'Clone which script? ',
               choices : fileIds,
             }]).then((answers: any) => {
+              answers.scriptId = answers.scriptId.substring(answers.scriptId.lastIndexOf('(') + 1,
+                                 answers.scriptId.length - 1);
               checkIfOnline();
               spinner.setSpinnerTitle(LOG.CLONING);
               saveProjectId(answers.scriptId);


ajlai:
@campionfellin it looks like we can just pass in an array of `{name:_, value_}` to avoid adding any parsing logic for `answers.scriptId`. See https://github.com/SBoudrias/Inquirer.js/#question

campionfellin:
@ajlai great find! Feel free to make a PR to fix that, or I can take a look at it in the next couple days.

PR #194, opened 2018-05-29
Fix small errors with clasp clone and clasp logs

campionfellin:
Thanks to @ajlai for finding the array of `{name:_,value:_}` in Inquirer.

Fixes #<issue_number_goes_here> (it's a good idea to open an issue first for discussion)

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -300,10 +300,10 @@ commander
             q: 'mimeType="application/vnd.google-apps.script"',
           });
           const files = data.files;
-          const fileIds:string[] = [];
+          const fileIds: Array<{ name: string; value: string; }> = [];
           if (files.length) {
             files.map((file: any) => {
-              fileIds.push(`${file.name}`.padEnd(20) + ` - (${file.id})`);
+              fileIds.push({name: `${file.name}`.padEnd(20) + ` - (${file.id})`, value: `${file.id}`});
             });
             await prompt([{
               type : 'list',
@@ -311,8 +311,6 @@ commander
               message : 'Clone which script? ',
               choices : fileIds,
             }]).then((answers: any) => {
-              answers.scriptId = answers.scriptId.substring(answers.scriptId.lastIndexOf('(') + 1,
-                                 answers.scriptId.length - 1);
               checkIfOnline();
               spinner.setSpinnerTitle(LOG.CLONING);
               saveProjectId(answers.scriptId);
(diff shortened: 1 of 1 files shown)


index.ts
```diff
@@ -300,19 +300,17 @@ commander
             q: 'mimeType="application/vnd.google-apps.script"',
           });
           const files = data.files;
-          const fileIds:string[] = [];
+          const fileIds: Array<{ name: string; value: string; }> = [];
           if (files.length) {
             files.map((file: any) => {
-              fileIds.push(`${file.name}`.padEnd(20) + ` - (${file.id})`);
+              fileIds.push({name: `${file.name}`.padEnd(20) + ` - (${file.id})`, value: `${file.id}`});
```
grant: I think this should be formatted on multiple lines.

index.ts
```diff
@@ -300,19 +300,17 @@ commander
             q: 'mimeType="application/vnd.google-apps.script"',
           });
           const files = data.files;
-          const fileIds:string[] = [];
+          const fileIds: Array<{ name: string; value: string; }> = [];
```
grant: Can you not do this?
```ts
const fileIds: {
  name: string;
  value: string;
}[] = [];
```

index.ts
grant:
```ts
fileIds.push({
  name: `${file.name.padEnd(20)} - (${file.id})`,
  value: file.id,
});
```

index.ts
campionfellin: `tslint` gives this error: `Array type using 'T[]' is forbidden for non-simple types. Use 'Array<T>' instead.`

campionfellin: https://github.com/google/clasp/blob/60393ef2f693f6f2cfa7b2b63cbbed1fa0f8438e/tslint.json#L5

index.ts
ajlai: why is this not just
```ts
            const fileIds = files.map((file) => {
              return {
                name: `${file.name.padEnd(20)} - (${file.id})`,
                value: file.id
              };
            });
``` 
?
Typescript can figure out the type here, no need to explicitly state it.

grant: 👍 This would be an even better refactor.

campionfellin: Perfect, this has now been fixed.

campionfellin:
I also noticed that `clasp clone` does not check if a `.clasp.json` file already exists in the directory. Do we want to do that?

grant:
Yes.
- `clasp clone` should fail if there is an existing `.clasp.json` file
- There should be a test case for this.

PR #198, opened 2018-05-30
Make running of tests easier

### LINKED ISSUE #197: Tests should be easier to run
## Expected Behavior

You download `clasp`, open up `tests/test.ts` and find and replace `describe.skip` with `describe` (to enable all tests) and then `sudo npm run build && npm run test`. All tests should pass with minimal setup.

## Actual Behavior

As it stands, there's a handful of tests that don't work without extra setup (namely: `clasp push`, `clasp deploy`, `clasp deployments`, `clasp version`)

  1. `git clone https://github.com/google/clasp && cd clasp`
  1. `sudo npm install`
  1.  Un-skip all the tests
  1. `sudo npm run build && npm run test`

## Specifications

  - Node version (`node -v`): 8.9.4
  - Version (`npm list -g | grep clasp`): 1.3.1
  - OS (Mac/Linux/Windows): MacOS

campionfellin:
This can help us make sure all tests actually pass and that we are increasing in code coverage.

Fixes #197

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -387,7 +387,7 @@ commander
                 nonIgnoredFilePaths.map((filePath: string) => {
                   console.error(`└─ ${filePath}`);
                 });
-                process.exit(-1);
+                process.exit(1);
               } else {
                 nonIgnoredFilePaths.map((filePath: string) => {
                   console.log(`└─ ${filePath}`);
--- a/tests/test.ts
+++ b/tests/test.ts
@@ -9,6 +9,9 @@ import { getScriptURL, getFileType, getAPIFileType,
 import * as path from 'path';
 import * as tmp from 'tmp';
 
+const TEST_CODE_JS = 'function test() { Logger.log(\'test\'); }';
+const TEST_JSON = '{"timeZone": "America/New_York"}';
+
 describe('Test help for each function', () => {
   it('should output help for run command', () => {
     const result = spawnSync(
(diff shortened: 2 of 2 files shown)


index.ts
```diff
@@ -387,7 +387,7 @@ commander
                 nonIgnoredFilePaths.map((filePath: string) => {
                   console.error(`└─ ${filePath}`);
                 });
-                process.exit(-1);
+                process.exit(1);
```
campionfellin: -1 is actually out of range for status codes, and really returns 255. Changed to 1 for consistency.

tests/test.ts
```diff
@@ -94,6 +94,9 @@ describe.skip('Test clasp pull function', () => {
 
 describe.skip('Test clasp push function', () => {
   it('should push local project correctly', () => {
+    fs.removeSync('.claspignore');
+    fs.writeFileSync('Code.js', ' ');
```
campionfellin: When using `fs.writeFileSync` we need to make sure to at least write a blank space (so `' '` rather than `''` so that the files can be found here. https://github.com/google/clasp/blob/2cd12253a9c77dcd8c021a4949110dc1b9057b41/index.ts#L80

tests/test.ts
```diff
@@ -165,31 +169,12 @@ describe.skip('Test clasp open function', () => {
   });
 });
 
-describe.skip('Test clasp clone function', () => {
```
campionfellin: This is moved below the `clasp deploy`, `clasp deployments`, and `clasp version` tests because this removes the `.clasp.json` file, which is needed for those tests.

tests/test.ts
grant: Use a constant string instead of a mysterious `' '`. Like:
`const TEST_CODE_JS = 'function test() { Logger.log('test'); }`;

PR #202, opened 2018-05-31
Load credentials is now async

### LINKED ISSUE #201: Async Credentials
## Expected Behavior

Loading credentials should be `async`. Part of #180.

```ts
async getAPICredentials();
const { scriptId, rootDir } = await getProjectSettings();
```

Rather than making a method like `getScript()` or `getDrive()`, we should load the credentials before calling the API methods. This way we don't need N number of `getAPI()` methods.

## Actual Behavior

```ts
getAPICredentials(async () => {
    const { scriptId, rootDir } = await getProjectSettings();
```

campionfellin:
Though it looks like a lot of changes, note that most of it is un-indenting the callbacks that were originally given to `getAPICredentials()`. It is now an `async` function, and any command that uses it has been changed to use `await getAPICredentials()` instead. 

Fixes #201 
Results:
```
  19 passing (55s)
  4 pending

--------------|----------|----------|----------|----------|-------------------|
File          |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
--------------|----------|----------|----------|----------|-------------------|
All files     |    56.37 |    43.12 |    60.95 |    62.89 |                   |
 clasp        |    96.55 |       50 |      100 |    96.55 |                   |
  index.js    |    96.55 |       50 |      100 |    96.55 |               187 |
 clasp/src    |     53.9 |    47.11 |    59.53 |    59.03 |                   |
  auth.js     |    28.07 |    31.91 |    27.27 |    31.82 |... 39,240,241,245 |
  commands.js |    56.74 |    49.11 |    60.22 |    59.08 |... 05,706,707,733 |
  files.js    |    67.11 |    55.14 |     93.1 |    78.57 |... 82,192,194,197 |
  utils.js    |    60.61 |    48.21 |    67.35 |    70.16 |... 43,268,269,294 |
 clasp/tests  |    61.25 |     9.38 |     66.1 |    72.64 |                   |
  test.js     |    61.25 |     9.38 |     66.1 |    72.64 |... 55,256,257,258 |
--------------|----------|----------|----------|----------|-------------------|
```

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -66,21 +66,19 @@ async function authorize(useLocalhost: boolean, writeToOwnKey: boolean) {
 /**
  * Loads the Apps Script API credentials for the CLI.
  * Required before every API call.
- * @param {Function} cb The callback
- * @param {boolean} isLocal If we should load local API credentials for this clasp project.
  */
-export function getAPICredentials(cb: (rc: ClaspSettings | void) => void) {
-    DOTFILE.RC_LOCAL.read().then((rc: ClaspSettings) => {
+export async function getAPICredentials() {
+    return DOTFILE.RC_LOCAL.read().then((rc: ClaspSettings) => {
       oauth2Client.setCredentials(rc);
-      cb(rc);
+      return rc;
     }).catch((err: any) => {
-      DOTFILE.RC.read().then((rc: ClaspSettings) => {
+      return DOTFILE.RC.read().then((rc: ClaspSettings) => {
         oauth2Client.setCredentials(rc);
-        cb(rc);
+        return rc;
       }).catch((err: any) => {
         console.error('Could not read API credentials. Error:');
         console.error(err);
-        process.exit(-1);
+        process.exit(1);
       });
     });
   }
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -68,17 +68,20 @@ async function authorize(useLocalhost: boolean, writeToOwnKey: boolean) {
  * Required before every API call.
  */
 export async function loadAPICredentials() {
-    return DOTFILE.RC_LOCAL.read().then((rc: ClaspSettings) => {
-      oauth2Client.setCredentials(rc);
-      return rc;
-    }).catch((err: any) => {
-      return DOTFILE.RC.read().then((rc: ClaspSettings) => {
+    return new Promise((resolve, reject) => {
+      DOTFILE.RC_LOCAL.read().then((rc: ClaspSettings) => {
         oauth2Client.setCredentials(rc);
-        return rc;
+        resolve();
       }).catch((err: any) => {
-        console.error('Could not read API credentials. Error:');
-        console.error(err);
-        process.exit(1);
+        DOTFILE.RC.read().then((rc: ClaspSettings) => {
+          oauth2Client.setCredentials(rc);
+          resolve();
+        }).catch((err: any) => {
+          console.error('Could not read API credentials. Error:');
+          console.error(err);
+          reject();
+          process.exit(1);
+        });
       });
     });
   }
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -68,22 +68,17 @@ async function authorize(useLocalhost: boolean, writeToOwnKey: boolean) {
  * Required before every API call.
  */
 export async function loadAPICredentials() {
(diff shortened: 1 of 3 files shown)


src/auth.ts
```diff
@@ -66,21 +66,19 @@ async function authorize(useLocalhost: boolean, writeToOwnKey: boolean) {
 /**
  * Loads the Apps Script API credentials for the CLI.
  * Required before every API call.
- * @param {Function} cb The callback
- * @param {boolean} isLocal If we should load local API credentials for this clasp project.
  */
-export function getAPICredentials(cb: (rc: ClaspSettings | void) => void) {
-    DOTFILE.RC_LOCAL.read().then((rc: ClaspSettings) => {
+export async function getAPICredentials() {
+    return DOTFILE.RC_LOCAL.read().then((rc: ClaspSettings) => {
```
campionfellin: One thing I want to note, is should we return the `rc` or not? It is never used, but maybe it will be in the future.

grant: Good comment. Had to think a bit.
If in doubt, don't expose variables/methods/values.
After `oauth2Client.setCredentials(rc);`, we don't need the credentials elsewhere so don't return the values.

In this case, remove `return rc;`
This Promise should resolve, but not return the `rc`.

src/auth.ts
```diff
@@ -66,21 +66,19 @@ async function authorize(useLocalhost: boolean, writeToOwnKey: boolean) {
 /**
  * Loads the Apps Script API credentials for the CLI.
  * Required before every API call.
- * @param {Function} cb The callback
- * @param {boolean} isLocal If we should load local API credentials for this clasp project.
  */
-export function getAPICredentials(cb: (rc: ClaspSettings | void) => void) {
-    DOTFILE.RC_LOCAL.read().then((rc: ClaspSettings) => {
+export async function getAPICredentials() {
```
grant: Now that we don't care about the Promise's return values (we still care about the Promise though), we can rename this to `loadAPICredentials`.

src/auth.ts
```diff
@@ -66,21 +66,22 @@ async function authorize(useLocalhost: boolean, writeToOwnKey: boolean) {
...
- * @param {boolean} isLocal If we should load local API credentials for this clasp project.
  */
-export function getAPICredentials(cb: (rc: ClaspSettings | void) => void) {
-    DOTFILE.RC_LOCAL.read().then((rc: ClaspSettings) => {
-      oauth2Client.setCredentials(rc);
-      cb(rc);
-    }).catch((err: any) => {
-      DOTFILE.RC.read().then((rc: ClaspSettings) => {
+export async function loadAPICredentials() {
+    return new Promise((resolve, reject) => {
```
campionfellin: Is this how we want it? Tests pass so it seems to work, just want to make sure this is the best way forward.

src/auth.ts
```diff
@@ -66,21 +66,22 @@ async function authorize(useLocalhost: boolean, writeToOwnKey: boolean) {
...
-        console.error(err);
-        process.exit(-1);
+        DOTFILE.RC.read().then((rc: ClaspSettings) => {
+          oauth2Client.setCredentials(rc);
+          resolve();
+        }).catch((err: any) => {
+          console.error('Could not read API credentials. Error:');
+          console.error(err);
+          reject();
+          process.exit(1);
```
campionfellin: Related to above comment, if we reject, then should we still exit after?

src/auth.ts
grant: No. `DOTFILE.RC_LOCAL.read()` already returns a Promise. There's no need to wrap a Promise around the Promise. Promises can be chained too, so we're okay to leave this as is. All we care about is loading `oauth2Client` before calling API methods, it really doesn't matter what happens after that.

You can try this in Chrome console:
```js
var a = new Promise();
function c() {
    return a.then(() => {}).catch(() => {});
}
c();
// Promise {<pending>}
```

src/auth.ts
grant: There's no need to reject a promise before a `process.exit()`.

I believe `resolve/reject` will just mark the function a certain way (fulfilled, pending, etc) before returning back to the caller. In this case it doesn't matter since we're killing the process and won't pop the function stack.

grant:
Thanks again!

PR #205, opened 2018-06-01
Generate README How To Section

### LINKED ISSUE #200: Generate README Doc
## Expected Behavior

The README methods should be generated from the TypeScript file.

## Actual Behavior

The README is slightly out of date, duplicates jsdoc comments, and must be hand edited.

## Solution

- Create a file called `generateDocs.ts`.
- Running `tsc generateDocs.ts && node generateDocs.js` should build the README without modifications needed.
  - This script should use the [TS Compiler API](https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API) over index.ts. A header and footer would be added to the README.
- A test should check that the current README file matches the latest generated README file.

grant:
Fixes #200 by introducing a `npm run docs` command.

Keeping the docs in sync with the JSDoc comments with the code is troublesome. In the past there have been instances where we update commands but don't update the README.

Our `index.ts` already does a good job of documenting the command line interface, so let's just ensure the comments are up-to-date there and run the generator to build part of the README.

Ideally in a later step we can add some code that actually modifies the README file.

### DIFF
--- a/docs.ts
+++ b/docs.ts
@@ -1,0 +1,79 @@
+/**
+ * Generates the How To section of the docs.
+ * Ensures the code for clasp commands is well documented.
+ * @todo Generate a list of commands for the beginning of the README.
+ */
+const fs = require('fs');
+const parseComments = require('parse-comments');
+const extract = require('extract-comments');
+const file = fs.readFileSync('./index.ts').toString();
+const ucfirst = require('ucfirst');
+
+// The README will be a concatenation of lines in this variable.
+let readme = [
+  '## How To...',
+];
+
+// Remove first line (#!/usr/bin/env node)
+const fileLines = file.split('\n');
+fileLines.splice(0, 1);
+const fileWithoutFirstLine = fileLines.join('\n');
+
+// Extract JSDoc comments out of our file.
+const comments = extract(fileWithoutFirstLine);
+for (const command of comments) {
+  // To use the parseComments module, complete the stripped comment.
+  const comment = `/*${command.raw}*/`;
+  const claspCommand = parseComments(comment)[0];
+  // Only print valid commands.
+  if (claspCommand && claspCommand.description && claspCommand.name) {
+    readme.push('');
+    readme.push(`### ${ucfirst(claspCommand.name)}`);
+    readme.push('');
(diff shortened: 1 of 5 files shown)


docs.ts
```diff
@@ -0,0 +1,79 @@
+/**
+ * Generates the How To section of the docs.
+ * Ensures the code for clasp commands is well documented.
+ * @todo Generate a list of commands for the beginning of the README.
+ */
+const fs = require('fs');
+const parseComments = require('parse-comments');
+const extract = require('extract-comments');
+const file = fs.readFileSync('./index.ts').toString();
+const ucfirst = require('ucfirst');
```
campionfellin: Not a big deal, but we're importing a one-liner:

```
'use strict';

module.exports = function ucfirst(s) {
  return s.charAt(0).toUpperCase() + s.substr(1);
};
```
from [here](https://github.com/deathcap/ucfirst/blob/master/ucfirst.js)

Can we just do this ourselves?

campionfellin:
@ajlai , do you want to make a PR with an example README that has been generated with your solution? I think that would be super awesome if it works!

ajlai:
@campionfellin here is a test branch (not fully implemented). https://github.com/ajlai/clasp/blob/autogenerate-readme/GENERATED-README.md 

The autogenerated section is here: https://github.com/ajlai/clasp/blob/autogenerate-readme/generate-readme/GENERATED-README.hbs#L49-L55 and triggered by `npm run generate-readme` (https://github.com/ajlai/clasp/blob/autogenerate-readme/package.json#L11)

campionfellin:
@ajlai I think that looks like it will work as well. Either solution is fine with me. @grant ?

docs.ts
grant: Sure, I'll update soon.

grant:
I'm fine with either solution. 
Just be sure to fine-tune jsdoc2md if we use that. I'm in favor of merging this PR and using `jsdoc2md` later since this PR also adds more details to the jsdoc comments.

campionfellin:
Approving for now. I'd like to see the `ucfirst` module removed in a future PR.

PR #206, opened 2018-06-01
Exit with non-zero status code

### LINKED ISSUE #159: Exit with an error on non-existing commands
## Expected Behavior

On running a mispelled command, or `help` (instead of the flag) or `list` (unimplemented), the program should print a message and exit with code > 0.

## Actual Behavior

Nothing is printed and the exit code is 0.

This is particularly confusing for `list`, as that was the first command I tried and empty output suggests it can't find any projects.

## Specifications

  - Node version (`node -v`): v10.0.0
  - Version (`npm list -g | grep clasp`): @google/clasp@1.1.5
  - OS (Mac/Linux/Windows): Mac

grant:
Fixes #159.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -82,6 +82,7 @@ export const push = async () => {
  */
 export const help = () => {
   commander.outputHelp();
+  process.exit(1);
 };
 
 /**


campionfellin:
Do we want to use `process.exit(1)` or `logError(...)` ?

Otherwise LGTM

grant:
We're not logging any error message here, so it wouldn't make sense.

campionfellin:
Sorry I must have missed this, so we merged and reverted this? Are we going to just use what was discussed here? https://github.com/google/clasp/issues/159#event-1657362437

grant:
I realized your comment meant put the `exit` in `logError` rather than here. It probably would be better there.

That said, it really doesn't matter and you could argue for both places.

campionfellin:
Up to you if you want to make another PR

Issue #78, opened 2018-03-29
Serve locally for testing


sindhus:
## Expected Behavior

I expect there to be an option to serve the script locally for easier testing. 

## Actual Behavior

There is no clasp command-line option to serve the Google Apps Script locally. 

## Specifications

  - Node version (`node -v`): v9.8.0
  - Version (`npm list | grep clasp`):

```
➜  ~ npm list | grep clasp
➜  ~ which clasp
/usr/local/bin/clasp
```
  - OS (Mac/Linux/Windows): Mac

grant:
Are you looking to run functions locally?
That's almost impossible since Apps Script binds to Google Services like Gmail, Slides, etc.

sqrrrl:
Probably best tackled outside clasp. Really this is asking for a local apps script runtime. It's (mostly) feasible but a very large project to create an emulator that can be run locally.

I do find https://www.npmjs.com/package/gas-local to be useful in tests. Still have to mock out whatever services you use, but at least gives you loading and module (err, lack of) behavior similar to apps script.

grant:
Going to close this in favor of #11.

PR #210, opened 2018-06-04
Typo on #078

kant:

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Logs the user in. Saves the client credentials to an rc file.
 
 ### Logout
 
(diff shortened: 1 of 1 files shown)


campionfellin:
Line 78 looks like it was generated from `index.ts` here: https://github.com/google/clasp/blob/d8a017dccfc8d08db20bcfbe5cbf1ced3e21c0a5/index.ts#L74

So it's probably best to change that and then re-generate the docs using `npm run docs`.

grant:
Can you modify https://github.com/google/clasp/blob/master/index.ts#L74 too?
Then this PR looks good, thanks!

Issue #204, opened 2018-05-31
CLASP RUN PRODUCES ERRORS


jasonjurotich:
(Please search [existing issues](https://github.com/google/clasp/issues) before creating a new one.)

## Expected Behavior
A simple function should run to add text to a cell in a Google Spreadsheet

## Actual Behavior
errors

## Steps to Reproduce the Problem

  1. clasp create [name]
  2. clasp clone [ScriptId] (This is to download the Code.js)
  3. add following code to Code.js by VIM editor:

function myFunction() {
 SpreadsheetApp.openById('1acJRcIGXIYa9YNgO8g3DTrgFRO6Iqk0wNJoxYpsust8').getRange('A1').setValue('set');
   }

4. save and close (in VIM = :x )
5. clasp push (make sure everything is the same)
6. clasp run myFunction

## Specifications
I am using the GCP with a VM instance with Ubuntu 18 (up to date) with Node 10.2 and @google/clasp@1.3.1

The errors are in the document attached. I tried the script in a normal manner after clasp open and it worked, so something is off with the clasp run. Maybe something with permissions?
[errorclasp.txt](https://github.com/google/clasp/files/2059492/errorclasp.txt)

grant:
@campionfellin Can you look at this?
It looks like we need to document `run` a bit better.

campionfellin:
Yep, I'll take a look in a little bit.

campionfellin:
So after some investigation this is what I've found:

If you follow the instructions in this PR: https://github.com/google/clasp/pull/172 you should be able to use `clasp run` for simple functions that don't need extra scoping access.

More complex functions (like yours, requiring access to Sheets) are not yet supported by `clasp` (yet 😉). By following the instructions outlined in the above PR, you should get a response like: 

```
  code: 401,
  errors:
   [ { message: 'Request is missing required authentication credential. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.',
       domain: 'global',
       reason: 'unauthorized' } ] }
```

Which indicates that you don't have authorization. Of course this makes sense, because when we logged in we didn't give it the right scopes (for Sheets). Go here: https://github.com/google/clasp/blob/8e1c3fbeca11e06d22554b183b5d43d1c8490808/src/auth.ts#L20

and add this scope `https://www.googleapis.com/auth/spreadsheets` to the list. Then `clasp logout` and `clasp login` and then run your function. You should get a response like:

```
[...]

PR #223, opened 2018-06-19
Change login flow (experimental)

campionfellin:
This is the flow (more or less):

`clasp login`: logs you in using the default `clasp` credentials. Saves `.clasprc.json` to your `~` directory.
`clasp login --creds`: logs you in using the file `credentials.json` in the directory that you are running the command in. Saves `.clasprc.json` to your current directory.
`clasp login --creds other_creds.json`: logs you in using the credentials in `other_creds.json`. Saves `.clasprc.json` to your current directory.

Your credentials file should look like this:

```
{"installed":{"client_id":"239267426989-275aglft6htcfsdj7t5d3csdlogchamh.apps.googleusercontent.com","project_id":"project-id-xxxxxxxxx","auth_uri":"https://accounts.google.com/o/oauth2/auth","token_uri":"https://accounts.google.com/o/oauth2/token","auth_provider_x509_cert_url":"https://www.googleapis.com/oauth2/v1/certs","client_secret":"xxxxxxxx","redirect_uris":["urn:ietf:wg:oauth:2.0:oob","http://localhost"]}}
```

Related to #204 


campionfellin:
Please test before merging.

campionfellin:
(also note that `npm run docs` did not work for some reason to update the docs with this)

marcosscriven:
@campionfellin - would this help with https://github.com/google/clasp/pull/28?

campionfellin:
For now, you still have to use the browser to confirm the login. What this does is allow you to authenticate with your own Oauth client, rather than the default for `clasp`. This helps with things like `clasp run` which need more permissions than the default.

campionfellin:
Will update README in separate PR.

src/auth.ts
```diff
@@ -52,11 +53,30 @@ export const drive = google.drive({
  * Requests authorization to manage Apps Script projects.
  * @param {boolean} useLocalhost True if a local HTTP server should be run
  *     to handle the auth response. False if manual entry used.
+ * @param {any} creds either boolean or string. If true (boolean), then --creds
+ *     was used without any input file. Otherwise creds is a string of the file
+ *     they entered.
  */
-async function authorize(useLocalhost: boolean, writeToOwnKey: boolean) {
+async function authorize(useLocalhost: boolean, creds: any) {
```
grant: Can we never have any?
`string|boolean`?

campionfellin: Typescript is not a fan 😔

```
[ts]
Argument of type 'string | false' is not assignable to parameter of type 'string | number | Buffer | URL'.
  Type 'false' is not assignable to type 'string | number | Buffer | URL'.
```

when trying to read the credentials file:

```
    const credentials = JSON.parse(fs.readFileSync(creds, 'utf8'));
```

campionfellin: The main reason for doing this is this line:

```
  creds = creds === true ? 'credentials.json' : creds;
```

and the reason for that was because if you put a default value in commander, it sets `options.creds` to that value whether or not the user entered in the `--creds` flag.

So I removed the default value from the commander side of things and use this line to set it to `credentials.json` by default.

src/auth.ts
```diff
@@ -52,11 +53,30 @@ export const drive = google.drive({
...
+  let ownCreds = false;
+  creds = creds === true ? 'credentials.json' : creds;
```
grant: Never write `=== true`
Just `creds ? creds : '...'`;

src/auth.ts
grant: Can you just do `string` and default `creds` outside the body (in the caller)?

There has to be some way to not use `any`.
This is an internal method too, so we have flexibility to our method signature.

src/auth.ts
campionfellin: Strings evaluate to true so when they enter `--creds otherCredsFile.json` then we would rename it to `credentials.json` unfortunately. I see the other comments and there is probably a way to do this without using this or `any` so I'll look closer at the `commander` documentation.

campionfellin:
Ok, after some testing with it, I think the best solution is to remove the default value and make it so if the user uses `--creds` they *must* supply a credentials file. The reasoning is this:

If we default it to `./credentials.json`, or anything really, when the user uses `clasp login` without specifying `--creds`, the `options.creds` variable is still set to `./credentials.json`. Thus, if they have that file in their directory, but want to login globally with the default (old) `clasp` credentials, they can't.

@grant what do you think?

grant:
Agreed, that makes sense. Let's not default.

campionfellin:
Ok, now here is what happens:

`clasp login` : works as it used to, using the default `clasp` credentials.
`clasp login --creds` : gives error `error: option '--creds <file>' argument missing`
`clasp login --creds creds.json` (when `creds.json` exists and is formatted properly) : logs in user with those credentials, saves `./.clasprc.json` 
`clasp login --creds creds.json` (when `creds.json` doesn't exist): `Credentials file not found.`
`clasp login --creds creds.json` (when `creds.json` exists, but not right format) : `Incorrect credentials file format.`

campionfellin:
Important to note when we update the readme:

If you use your own credentials from GCP, you must enable both the Apps Script API and Drive API.

grant:
It would be cool if you added `@example` with the comment you wrote above, so it lasts.

campionfellin:
Added simple examples, I think the more detailed "what happens if I do this..." should be in the README under troubleshooting.

grant:
I'll try to release this week with these changes.

Issue #217, opened 2018-06-14
Enable Advanced Services for Apps Script projects via CLI


grant:
## Expected Behavior

To use Google APIs (Advanced Services) in Apps Script, you must enable APIs in the online editor at `script.google.com > Resources > Advanced Google Services...`

You should be able to setup a project without having to go to the editor. I propose a function to enable APIs in `clasp` like the UI.

### Commands

```sh
clasp apis list
clasp apis enable "drive.googleapis.com"
clasp apis disable "drive.googleapis.com"
```

- `list` would list all Google API `serviceId`s
- `enable`/`disable` a API for the Cloud project

Eventually autocomplete:

```sh
clasp apis enable # select from list of all Google APIs that aren't enabled
clasp apis disable # select from list of enabled APIs in the manifest
```

### API

These commands would use the `serviceuser.projects.services` API. It would require a personal Cloud project.
`projects/project-id-0486239153690933627/services/drive.googleapis.com`

**EDIT:** User [serviceusage](https://developers.google.com/apis-explorer/#p/serviceusage/v1/) not serviceuser since that's the newer API.

- List Services API: https://developers.google.com/apis-explorer/#p/discovery/v1/discovery.apis.list
- Enable Service API: https://developers.google.com/apis-explorer/#p/serviceuser/v1/serviceuser.projects.services.enable
- Disable Service API: https://developers.google.com/apis-explorer/#p/serviceuser/v1/serviceuser.projects.services.disable

### Result

The result of enabling an advanced service via the UI is:

1. Enabling the API for your Cloud Console project
1. Adding some data to your `appsscript.json`.

```js
  "dependencies": {
    "enabledAdvancedServices": [{
      "userSymbol": "AdSenseOrWhatever",
      "serviceId": "adsense",
      "version": "v1.4"
    }]
  }
```

## Actual Behavior

Users have to go to script.google.com to enable advanced services.

rdoursenaud:
The actual behavior is even more annoying since it disables the advanced services on `clasp push`!

grant:
@rdoursenaud Advanced service aren't disabled on `push`. You need to `pull` after you enable an advanced service, because it changes the `appsscript.json` file. LMK if you're still seeing an error.

I should document this more.

grant:
No worries, it's not intuitive (see #171).
I probably should add a warning if you try to `push` before `pull`ing like git.

rdoursenaud:
[...]

PR #228, opened 2018-06-21
clasp apis <subcommand>

campionfellin:
Starts work on `clasp apis <subcommand>` with `clasp apis list`.

Note, we should find a definitive list of APIs that users can enable before merging this PR. I found this: https://developers.google.com/identity/protocols/googlescopes

but that is too many to list for the user. Any ideas?

Also adds tests.

Works on  #217 

- [ ] Appropriate changes to README are included in PR. (no changes until it is done)


grant:
> Note, we should find a definitive list of APIs that users can enable before merging this PR

So there's an API for listing Google APIs (meta) (I updated the issue):

https://developers.google.com/apis-explorer/#p/discovery/v1/discovery.apis.list

---

For `apis list`, we want to use `name` and `id`.
Later on we can consider an option to add/remove these to the manifest, but I'd say let's just focus on `list`, `enable`, `disable`.

---

Maybe we should say `add` and `remove` like `yarn` if we're fully adding the ability to call the API from the single command?

campionfellin:
Continuing the conversation here, is there a canonical list of APIs we want? There would be much too many to show the user if we did all.

grant:
Can we just list the 198 APIs with 1 API per line?

Or we can create a list derived from the `Advanced Google Services` UI:

```
AdSense Management API		
Admin Directory API		
Admin Reports API		
BigQuery API		
Calendar API		
Content API for Shopping		
ShoppingContent
DCM/DFA Reporting And Trafficking API		
Drive API		
Enterprise Apps Reseller API		
Enterprise License Manager API		
Fusion Tables API		
G Suite Activity API		
Gmail API		
Google Analytics API		
Google Analytics Reporting API		
Google Classroom API		
Google Mirror API		
Google Sheets API		
Google Slides API		
Google+ API		
Google+ Domains API		
Groups Migration API		
Groups Settings API		
Tag Manager API		
Tasks API		
URL Shortener API		
YouTube Analytics API		
YouTube Data API		
Youtube Content ID API		
```

campionfellin:
I think as a user a shorter list would be better...

grant:
Well, let's be sure to implement a solution that doesn't restrict users to the apis  available at the time of installing clasp. Some users will never upgrade clasp and may get different results if we manually keep a list of supported APIs.

I think listing all of the APIs for `list`, and using vorpal autocomplete for enabling APIs would be a good solution.

campionfellin:
Sounds good, will do that then.

campionfellin:
Sample of what it looks like:

screen shot 2018-06-26 at 5 08 06 pm

campionfellin:
Reduced it from 198 APIs to 134 APIs by using `preferred: true` (https://developers.google.com/discovery/v1/reference/apis/list)

src/auth.ts
```diff
@@ -47,6 +48,9 @@ export const drive = google.drive({
   version: 'v3',
   auth: oauth2Client,
 }) as Drive;
+export const discovery = google.discovery({
```
campionfellin: @grant not sure if this should be in the `auth` file, as you don't need to authenticate to use it. However, it made sense at the time to put it there.

src/auth.ts
```diff
@@ -44,6 +46,7 @@ google.options({
 export const script = google.script({version: 'v1'}) as Script;
 export const logger = google.logging({version: 'v2'}) as Logging;
 export const drive = google.drive({version: 'v3'}) as Drive;
+export const discovery = google.discovery({version: 'v1'}) as Discovery;
```
campionfellin: @grant not sure if this should be in the auth file, as you don't need to authenticate to use it. However, it made sense at the time to put it there.

campionfellin:
Build is breaking because of this: (https://github.com/google/clasp/blob/69b8dd298359922631f11a1b219ae39596c513f4/src/auth.ts#L39) looking for a solution...

src/commands.ts
```diff
@@ -525,3 +525,41 @@ export const openCmd = async (scriptId: any) => {
...
+  const {data} = await discovery.apis.list({
+    preferred: true,
+  });
+  data.items.forEach((api) => {
+    console.log(`${api.name.padEnd(25)} - ${api.id.padEnd(30)}`);
+  });
+};
+
+/**
+ * Acts as a router to apis subcommands
```
grant: Expand this comment.

src/commands.ts
```diff
@@ -525,3 +525,41 @@ export const openCmd = async (scriptId: any) => {
     process.exit(0);
   }
 };
+
+/**
+ * Lists available apis for the user.
+ */
+const apisList = async () => {
```
grant: Decrease the scope of this method by inlining it inside the `apis` method. You can keep it as a function, but inside that method.

```ts
export const apis = async () => {
  const list = async () => {};

```

This way,
- No other method in this file can use `apisList`. (No TS autocompletion suggestions either)
- We can simplify the name of the name of the method from `apisList` to `list` since `list` is part of the `apis` closure.

campionfellin:
@grant I just tried `null`, `undefined`, and `''`, and none of them worked. May need to basically revert https://github.com/google/clasp/pull/237

grant:
If reverting #237 solves the issue, then I'm fine if this PR does that. We can probably keep the clients at 1 line though.

campionfellin:
Yep, and I think as far as needing authentication to enable APIs we wouldn't be using the `discovery` client for that anyway.

grant:
Well, I'm not sure what scopes are needed for enabling an API.
I guess we should re-auth the user with scopes for enabling/disabling if they try to do those functions? aka incremental auth.

For now, I'd just focus on making the functions work and give all scopes.

grant:
Scratch the comments about enable/disable. This PR can focus on `list`.

src/commands.ts
```diff
@@ -525,3 +525,41 @@ export const openCmd = async (scriptId: any) => {
...
+    console.log(`${api.name.padEnd(25)} - ${api.id.padEnd(30)}`);
+  });
+};
+
+/**
+ * Acts as a router to apis subcommands
+ */
+export const apis = async () => {
+  const subcommand = process.argv[3];
+  switch (subcommand) {
```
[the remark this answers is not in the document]
campionfellin: Done.

index.ts
```diff
@@ -275,6 +276,16 @@ commander
   .description('Run a function in your Apps Scripts project')
   .action(run);
 
+/**
+ * @name apis
```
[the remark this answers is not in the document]
campionfellin: Done.

src/commands.ts
campionfellin: Done.

src/commands.ts
campionfellin: Done.

campionfellin:
Fails with NodeJS 6/7, I think because of using `forEach()`... (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach)

src/commands.ts
```diff
@@ -525,3 +525,31 @@ export const openCmd = async (scriptId: any) => {
...
+ * Calls functions for list, enable, or disable
+ * Otherwise returns an error of command not supported
+ */
+export const apis = async () => {
+  const list = async () => {
+    await checkIfOnline();
+    const {data} = await discovery.apis.list({
+      preferred: true,
+    });
+    data.items.forEach((api) => {
```
grant: Never write `forEach`, just use `for`:
```ts
for (let api of data.items) {
```

There are articles like these:
https://coderwall.com/p/kvzbpa/don-t-use-array-foreach-use-for-instead

src/commands.ts
```diff
@@ -525,3 +525,31 @@ export const openCmd = async (scriptId: any) => {
...
+      console.log(`${api.name.padEnd(25)} - ${api.id.padEnd(30)}`);
+    });
+  };
+  const subcommand: string = process.argv[3]; // clasp apis list => "list"
+  const command: any = {
```
grant: No `any`. Use a map.
I think it's like this:
`var x: {[key: KeyType]: ValueType} = {}`

src/commands.ts
campionfellin: Oh I had no idea! Thanks for the information. There's a couple other places where we find `forEach()` (in `clasp status`) that should be changed in a separate PR.

src/commands.ts
grant: Or maybe `Map<string, string>`

grant:
Can you test on those versions of Node and debug the problem?:
```sh
sudo npm i -g n
sudo n 6.0.0
sudo n 7.0.0
sudo n # choose the version you want to test
npm run test
```

campionfellin:
I believe it is `padEnd()` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padEnd)

Might be time to go back to using the `.clasprc.json.enc` to run the full test suite... Though this means external PRs would need to be manually re-built with Travis. (external PRs can't decrypt encrypted files).

grant:
Ah yes, `padEnd` probably doesn't work on older versions of Node.

Do you want to add a polyfill like `pad-end` for this PR and we can create an issue to replace it everywhere?

I'm guessing `padEnd` was only supported since version 8+ (2017-05-30)

PR #229, opened 2018-06-21
Add clasp -v and clasp --version

campionfellin:
May be helpful for debugging use issues.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -284,6 +284,12 @@ commander
   .action(help);
 
 /**
+ * Displays clasp version
+ */
+commander
+.version('1.3.1', '-v, --version');
+
+/**
  * All other commands are given a help message.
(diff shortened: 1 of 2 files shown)


index.ts
```diff
@@ -283,6 +283,12 @@ commander
   .description('Display help')
   .action(help);
 
+/**
+ * Displays clasp version
+ */
+commander
+.version('1.3.1', '-v, --version');
```
grant: This number will never be kept in sync with the actual latest version. Can we bump versions in one place?

https://github.com/google/clasp/blob/master/package.json#L3
https://github.com/tj/commander.js/issues/48#issuecomment-49087390

campionfellin: Good call, thanks 👍

PR #233, opened 2018-06-25
Improve .claspignore documentation.

richardkmichael:
Don't suggest `.claspignore` is like `.gitignore`, because it is not.

In fact, differences between the files are specifically mentioned; make them more prominent to avoid confusion.


grant:
Thanks! I'll merge this once the CLA is signed.

richardkmichael:
Curious, I signed the CLA right after I sent the PR... did I make a mistake somewhere?

PR #232, opened 2018-06-23
Modify tests to avoid spawnSync behavior on Windows

wilt00:
Fixes #231

- [X] `npm run test` succeeds. (!)
- [X] `npm run lint` succeeds.
- [X] Appropriate changes to README are included in PR. (N/A)

PR #235, opened 2018-06-26
Join commands with ampersands in package.json

wilt00:
As mentioned in #232. Additionally, the trailing semicolons seemed to be causing problems, so I've removed those.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,8 @@
   "description": "Develop Apps Script Projects locally",
   "main": "index.js",
   "scripts": {
-    "build": "tsc --project tsconfig.json; npm i -g --loglevel=error;",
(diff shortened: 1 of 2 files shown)


wilt00:
While testing `npm run build-fresh`, I noticed that `open` has a critical security hole (https://nodesecurity.io/advisories/663) and no updates for 4 years. Would you consider a pull request to replace it with e.g. `opn`?

grant:
Can you pull `master` again? I think the
`const CLASP = (os.type() === 'Windows_NT') ? 'clasp.cmd' : 'clasp';`
should be in master now. So this PR should be small.

For `opn`, can you open a new issue to discuss that?

PR #237, opened 2018-06-26
Unify oauthclient usage

campionfellin:
While doing OAuth work today, noticed we could condense this into one call.


campionfellin:
Yes, good call. Will merge when green.

PR #239, opened 2018-06-26
Fix a couple of tests

campionfellin:
Changed `logError(...)` in a couple places back to `console.error(...)` because I noticed this test was failing:

https://github.com/google/clasp/blob/e3866dc6df21193b0412ab98e8296905f5b1386f/tests/test.ts#L117

Fixed the other test by removing the `.clasp.json` file before testing, otherwise the response would be `Project file .clasp.json already exists.`. Then re-write `.clasp.json` for use in further tests.

### DIFF
--- a/src/files.ts
+++ b/src/files.ts
@@ -195,13 +195,13 @@ export async function pushFiles() {
         }, {}, (error: any) => {
           spinner.stop(true);
           if (error) {
-            logError(null, LOG.PUSH_FAILURE);
+            console.error(LOG.PUSH_FAILURE);
             error.errors.map((err: any) => {
-              logError(null, err.message);
+              console.error(err.message);
             });
-            logError(null, LOG.FILES_TO_PUSH);
+            console.error(LOG.FILES_TO_PUSH);
             nonIgnoredFilePaths.map((filePath: string) => {
-              logError(null, `└─ ${filePath}`);
+              console.error(`└─ ${filePath}`);
             });
             process.exit(1);
           } else {
--- a/src/files.ts
+++ b/src/files.ts
@@ -194,6 +194,10 @@ export async function pushFiles() {
           resource: { files },
         }, {}, (error: any) => {
           spinner.stop(true);
+          // In the following code, we favor console.error()
+          // over logError() because logError() exits, whereas
+          // we want to log multiple lines of messages, and
+          // eventually exit after logging everything.
           if (error) {
             console.error(LOG.PUSH_FAILURE);
             error.errors.map((err: any) => {
--- a/tests/test.ts
+++ b/tests/test.ts
@@ -82,11 +82,14 @@ describe.skip('Test clasp clone <scriptId> function', () => {
     expect(result.status).to.equal(0);
   });
   it('should give an error on a non-existing project', () => {
+    const settings = JSON.parse(fs.readFileSync('.clasp.json', 'utf8'));
+    fs.removeSync('.clasp.json');
     const result = spawnSync(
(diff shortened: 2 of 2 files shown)


tests/test.ts
```diff
@@ -82,11 +82,14 @@ describe.skip('Test clasp clone <scriptId> function', () => {
     expect(result.status).to.equal(0);
   });
   it('should give an error on a non-existing project', () => {
+    const settings = JSON.parse(fs.readFileSync('.clasp.json', 'utf8'));
+    fs.removeSync('.clasp.json');
```
grant: Why are we removing this file? Where was it created? We should delete the file in the same test it was created so that the tests are isolated.

src/files.ts
```diff
@@ -195,13 +195,13 @@ export async function pushFiles() {
         }, {}, (error: any) => {
           spinner.stop(true);
           if (error) {
-            logError(null, LOG.PUSH_FAILURE);
+            console.error(LOG.PUSH_FAILURE);
```
grant: Can we remove this change?
If we're going to change `logError`, we should do it everywhere and in a different PR.

tests/test.ts
campionfellin: Yeah, so we're removing it because otherwise you'll get the `Project file .clasp.json already exists.` error, whereas we are really targeting the error that says `> Did you provide the correct scriptId?`. We should probably have an additional test for when a `.clasp.json` already exists.

We re-write it at the end of the test so that the following test, for `clasp pull` works. We could definitely separate things though if that's what is better.

src/files.ts
campionfellin: Hey, so we're not changing it everywhere, just in this specific case:

```
          if (error) {
            console.error(LOG.PUSH_FAILURE);
            error.errors.map((err: any) => {
              console.error(err.message);
            });
            console.error(LOG.FILES_TO_PUSH);
            nonIgnoredFilePaths.map((filePath: string) => {
              console.error(`└─ ${filePath}`);
            });
            process.exit(1);
```
Since `logError` ends with `process.exit(1)` we never reach the next logs, so if there is an error on this command, it only shows:

`Push failed. Errors:`

And no errors, because it exits. I think this is an isolated enough problem that we don't need any more changes. One thing we could do is have `logError` accept a boolean of whether or not to exit.

grant: OK. Please comment this reason in the code.
We can change the `logError` logic later so we exit properly everywhere.

tests/test.ts
grant: OK.
Can you please include a detailed comment for these points:
- In the file comment, state `.clasp.json` exists for this test suite as a global file.
- In this test, state we are temporarily removing `.clasp.json` to test `clasp clone` fails when the project doesn't have this file.

campionfellin: So it's not really global, it's created as the tests go through. That's why they're in a specific order. It's definitely not ideal, each test should be independent of another, but it's what we have for now.

campionfellin: To clarify, it is created in a previous test, which tests `clasp create <title>`. Then we use the `.clasp.json` from that created script in the following tests.

src/files.ts
campionfellin: Done.

tests/test.ts
grant: OK. We'll really need to fix this later, as we can't have dependencies between test (where reordering the tests will fail).

campionfellin: @grant yeah I know, it's tough though because many tests need a real `scriptId` for them to work 😞

PR #242, opened 2018-06-26
Fix issue with opn library

campionfellin:
Noticed this while running tests, the `open(...)` command doesn't return an exit status on MacOS (see: https://github.com/sindresorhus/opn/issues/92). 

Thus the test for `clasp open` was hanging up even though it successfully opened.

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -124,6 +124,7 @@ async function authorizeWithLocalhost() {
     const authUrl = client.generateAuthUrl(oauth2ClientAuthUrlOpts);
     console.log(LOG.AUTHORIZE(authUrl));
     open(authUrl);
+    process.exit(0);
   });
   server.close();
   return (await client.getToken(authCode)).tokens;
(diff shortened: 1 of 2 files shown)


DanielRuf:
Just as sidenot. I tried the same in one of my projects.
Calling `process.exit` does not trigger `opn` on Windows and immediately closes the application.

DanielRuf:
`opn` returns a promise so something like

```
await opn('https://www.ueberbit.de/jobs/' + job)
process.exit(0)
```

or

```
opn('https://www.ueberbit.de/jobs/' + job)
setTimeout(()=>process.exit(0),300)
```

should do the trick.

As the promise does not resolve only option 2 works on both platforms.

DanielRuf:
Try opn(target, {wait: false}).

DanielRuf:
Try `opn(target, {wait: false})`.

grant:
@campionfellin Could we `await` opn?

DanielRuf:
> @campionfellin Could we await opn?

This will not help here as it will never resolve.

DanielRuf:
> As the promise does not resolve only option 2 works on both platforms.

But https://github.com/google/clasp/pull/242#issuecomment-410847584 is probably better.

grant:
OK, let's
`opn(target, {wait: false})`
@DanielRuf Can you open a new GitHub issue for this?

PR #250, opened 2018-07-02
Re encrypt

campionfellin:
Re-encrypted `.clasprc.json` and update keys.


campionfellin:
Looks like it finished, with the build error being only for Node 7 (https://travis-ci.org/google/clasp/jobs/399342926), which will be fixed in #251

PR #253, opened 2018-07-03
Better error logging for quota limit

campionfellin:
Drive API seems to have a daily limit, which is hit a lot with the full tests, (`clasp create` is called 10 times for every PR).

This changes the error message from `Error creating script.` to `Rate limit exceeded. Check quota.` so that we know when this happens if builds fail.

Looks like I can request a higher quota here: https://support.google.com/code/contact/drive_quota?

but it's unclear what to put for Project ID in the form..


grant:
I've requested 50 QPS and 100000 QPD with my Google email.
The Apps Script API is whitelisted for this project, but not the Drive API.
I'll see what happens.

campionfellin:
Sounds good, if you want to switch over the `.clasp.json.enc` file to your account let me know and I can help with that.

PR #254, opened 2018-07-03
Add note about Travis quota limits

grant:
The Travis build occasionally fails because of API quotas.
Let's note that.


campionfellin:
LGTM. I think a helpful change might be something like this in the `clasp create` test:

```sh
if (result.stderr == 'quota limit`) {
    isPR=true;
}
```

(psuedo-code) but essentially in the first test, if we hit the limit then pretend like it is a PR and skip all the other tests. Thus we wouldn't be failing, but rather skipping.

PR #259, opened 2018-07-17
Add note about tsc dependency

grant:
Tests require `tsc` installed. State that.
EDIT: Do not merge. I believe the tests instructions issue is I was facing is not this, but simply not building clasp before testing.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -8,6 +8,8 @@ To run tests yourself, follow the __Local Configuration__ instructions.
 
 ## Local Configuration
 
+> Note: Requires `tsc` installed `npm install -g typescript`
+
 1. Make sure you are logged in (`clasp login`).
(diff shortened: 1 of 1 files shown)


ajlai:
Isn't the `tsc` dependency already satisfied in devDependencies here https://github.com/google/clasp/blob/master/package.json#L95 ? I was able to run `npm test` without having typescript globally installed.

grant:
@ajlai Thanks. That seems true. I'm trying to follow the instructions for https://github.com/google/clasp/tree/master/tests#local-configuration and I think I just forget to build (`sudo npm run build-fresh`).

I'll have to change this PR to update the test instructions so you can just copy-paste the setup.

grant:
I don't think this change is needed. Closing.

PR #269, opened 2018-07-28
Use .clasp.json file for each test

### LINKED ISSUE #265: Clasp tests should still work when clasp create fails
### Problem 
When `clasp create` fails, with error `Rate limit exceeded. Check quota.`, a `.clasp.json` file is not created, and thus subsequent tests fail. See `clasp create` test here: https://github.com/google/clasp/blob/0a7ac1366acfb8db91a9600fad9cee675ecce535/tests/test.ts#L61

This goes towards the long term goal of each test being independent of others.

### Proposed Solutions

If the `clasp create` test gets hit with the rate limit, then just `this.skip()` that test specifically. Also either:

#### 1
Create a `.clasp.json` file (one-time) and upload it to the `/tests` folder. Use this for the tests.
#### 2
 Save a `scriptId` in the Travis environmental variables, and use `fs.writeFile(...)` to make a `.clasp.json` file.

---

By doing this, we should be more resilient to build failures and the build should always pass (with varying degrees of coverage).

Note: I don't think that `scriptId`s are "sensitive", but if so, we can do the encrypted Travis file, like with the `.clasprc.json`

campionfellin:
This makes tests more atomic by giving each one their own `.clasp.json` file, rather than carrying one through from the `clasp create` test.

Also skips the `clasp create` test for now. In a separate PR, I will change to only skip if it fails.

Note: in the `.travis.yml` file, the `SCRIPT_ID` needs to be updated to one from @grant account, rather than mine.

Fixes #265

### DIFF
--- a/tests/test.ts
+++ b/tests/test.ts
@@ -89,9 +89,7 @@ describe('Test clasp create function', () => {
     );
     expect(result.stderr).to.contain('Project file (.clasp.json) already exists.');
   });
-  after(() => {
-    fs.removeSync('.clasp.json');
-  });
+  after(cleanup);
 });
 
 describe.skip('Test clasp create <title> function', () => {
--- a/tests/test.ts
+++ b/tests/test.ts
@@ -15,6 +15,14 @@ const CLASP_SETTINGS: string = JSON.stringify({
   scriptId: process.env.SCRIPT_ID,
 });
 
+const cleanup = () => {
+  fs.removeSync('.clasp.json');
+};
+
+const setup = () => {
+  fs.writeFileSync('.clasp.json', CLASP_SETTINGS);
+};
+
 describe('Test --help for each function', () => {
   const expectHelp = (command: string, expected: string) => {
     const result = spawnSync(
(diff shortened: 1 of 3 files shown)


tests/test.ts
```diff
@@ -78,9 +81,12 @@ describe('Test clasp create function', () => {
     );
     expect(result.stderr).to.contain('Project file (.clasp.json) already exists.');
   });
+  after(() => {
```
grant: Can you try extracting this after method into a common function to reduce duplicated code?

tests/test.ts
```diff
@@ -101,6 +107,7 @@ describe('Test clasp clone <scriptId> function', () => {
     if (isPR !== 'false') {
       this.skip();
     }
+    fs.writeFileSync('.clasp.json', CLASP_SETTINGS);
```
grant: Can you add a comment?

grant: Can you extract this line (`fs.writeFileSync`) to remove duplicate code/strings.

campionfellin:
@grant these have now been extracted into the 'setup' and 'cleanup` functions at the top of the file.

Issue #266, opened 2018-07-26
Clasp tests should hit every ERROR/LOG message


campionfellin:
To increase coverage we should try to find and hit every ERROR message:

https://github.com/google/clasp/blob/0a7ac1366acfb8db91a9600fad9cee675ecce535/src/utils.ts#L82-L114

with the tests.

Additionally, we should probably hit all the log messages as well:

https://github.com/google/clasp/blob/0a7ac1366acfb8db91a9600fad9cee675ecce535/src/utils.ts#L117-L155

grant:
I think this issue is nice to have, but isn't necessary.

Having 100% code coverage will forever be a goal of this project. In reality, other issues are more important and this project doesn't have enough staffing (just me part time) to maintain 100% coverage. Feel free to add PRs with this issue attached though.

PR #277, opened 2018-08-02
Add more tests for when running commands logged out

campionfellin:
These tests don't add a ton of substantiative testing, but do point out a lot of things we may want to address in the future. See comments.

Works on #266 

### DIFF
--- a/tests/test.ts
+++ b/tests/test.ts
@@ -5,7 +5,7 @@ import * as fs from 'fs-extra';
 import { describe, it } from 'mocha';
 import * as tmp from 'tmp';
 import { getFileType } from './../src/files';
-import { getAPIFileType, getScriptURL, saveProjectId } from './../src/utils.js';
+import { getAPIFileType, getScriptURL, saveProjectId, ERROR } from './../src/utils.js';
 const { spawnSync } = require('child_process');
 const TEST_CODE_JS = 'function test() { Logger.log(\'test\'); }';
 const TEST_JSON = '{"timeZone": "America/New_York"}';
@@ -14,6 +14,7 @@ const isPR = process.env.TRAVIS_PULL_REQUEST;
 const CLASP_SETTINGS: string = JSON.stringify({
   scriptId: process.env.SCRIPT_ID,
 });
+const CLASP_USAGE = 'Usage: clasp <command> [options]';
 
 const cleanup = () => {
   fs.removeSync('.clasp.json');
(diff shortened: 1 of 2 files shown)


tests/test.ts
```diff
@@ -429,3 +429,139 @@ describe('Test clasp logout function', () => {
     expect(dotExists).to.equal(false);
   });
 });
+
+describe('Test clasp help function (in general)', () => {
+  it('should show help correctly', () => {
+    const result = spawnSync(
+      CLASP, ['help'], { encoding: 'utf8' },
+    );
+    expect(result.stdout).to.contain('Usage: clasp clasp <command> [options]');
```
campionfellin: We don't need `${PROJECT_NAME}` here: https://github.com/google/clasp/blob/87b8de5ed6b31dfbbf145ba121b41ec18404d45d/src/index.ts#L57

Commander will put it in by itself. This is why we get `usage: clasp clasp` (twice)

tests/test.ts
```diff
@@ -429,3 +429,139 @@ describe('Test clasp logout function', () => {
...
+  });
+});
+
+describe('Test all functions while logged out', () => {
+  it('should list error correctly', () => {
+    const result = spawnSync(
+      CLASP, ['list'], { encoding: 'utf8' },
+    );
+    expect(result.status).to.equal(1);
+    expect(result.stderr).to.contain('Could not read API credentials. Error:');
```
campionfellin: This error comes from here:

https://github.com/google/clasp/blob/87b8de5ed6b31dfbbf145ba121b41ec18404d45d/src/auth.ts#L91

Why not put it in this?

https://github.com/google/clasp/blob/87b8de5ed6b31dfbbf145ba121b41ec18404d45d/src/utils.ts#L82

Then we can import it as `ERROR.NO_CREDENTIALS` or something like that.

tests/test.ts
```diff
@@ -429,3 +429,139 @@ describe('Test clasp logout function', () => {
...
+  });
+  it('should pull error correctly', () => {
+    const result = spawnSync(
+      CLASP, ['pull'], { encoding: 'utf8' },
+    );
+    expect(result.status).to.equal(1);
+    expect(result.stderr).to.contain('No .clasp.json settings found. `create` or `clone` a project first.');
```
campionfellin: Some commands are inconsistent of the order they check for `.clasp.json` vs `.clasprs.json`. This command looks for `.clasp.json` first:

https://github.com/google/clasp/blob/87b8de5ed6b31dfbbf145ba121b41ec18404d45d/src/commands.ts#L34

Before looking for credentials in the `fetchProject(...)` method that it calls here:

https://github.com/google/clasp/blob/87b8de5ed6b31dfbbf145ba121b41ec18404d45d/src/files.ts#L183

Should probably make the order of checking for things more consistent.

tests/test.ts
grant: Yes, please!

tests/test.ts
grant: Can we fix this? It seems like these tests were written to match the output.

Ideally, we write the expected output in tests, and then fix the program to match the desired output (`Usage: clasp <command> [options]`)

tests/test.ts
```diff
@@ -429,3 +429,139 @@ describe('Test clasp logout function', () => {
...
+      CLASP, ['help'], { encoding: 'utf8' },
+    );
+    expect(result.stdout).to.contain('Usage: clasp clasp <command> [options]');
+    expect(result.status).to.equal(1);
+  });
+  it('should show help correctly', () => {
+    const result = spawnSync(
+      CLASP, ['--help'], { encoding: 'utf8' },
+    );
+    expect(result.stdout).to.contain('Usage: clasp clasp <command> [options]');
```
grant: Create a `const` for `'Usage: clasp clasp <command> [options]'`

tests/test.ts
```diff
@@ -429,3 +429,139 @@ describe('Test clasp logout function', () => {
...
+});
+describe('Test clasp version function', () => {
+  it('should show version correctly', () => {
+    const result = spawnSync(
+      CLASP, ['--version'], { encoding: 'utf8' },
+    );
+    expect(result.stdout).to.contain(require('./../package.json').version);
+    expect(result.status).to.equal(0);
+  });
+  it('should show help correctly', () => {
```
grant: Fix the description.

tests/test.ts
grant: Let's create a separate issue and write down the order in which we should do things in a dev README comment like "The structure of a `clasp` command".

tests/test.ts
```diff
@@ -429,3 +429,139 @@ describe('Test clasp logout function', () => {
...
+  it('should show version correctly', () => {
+    const result = spawnSync(
+      CLASP, ['unknown'], { encoding: 'utf8' },
+    );
+    expect(result.stderr).to.contain('🤔  Unknown command "unknown"');
+    expect(result.status).to.equal(1);
+  });
+});
+
+describe('Test all functions while logged out', () => {
```
grant: This is a lot of repeated code. Can you refactor the code so that each test is 1 line?

tests/test.ts
grant: You can put a comment to the inconsistent ordering here too.

tests/test.ts
campionfellin: Yes, will make separate PR. Just didn't want a failing build

tests/test.ts
grant: > "should show help correctly"

This really "should show the clasp version".

tests/test.ts
```diff
@@ -447,3 +448,112 @@ describe('Test clasp logout function', () => {
...
+    // Should be ERROR.NO_CREDENTIALS
+    // see: https://github.com/google/clasp/issues/278
+    expect(result.stderr).to.contain(ERROR.SCRIPT_ID_DNE);
+  });
+  it('should redeploy error correctly', () => {
+    const result = spawnSync(
+      CLASP, ['redeploy', '1234'], { encoding: 'utf8' },
+    );
+    expect(result.status).to.equal(1);
+    expect(result.stderr).to.contain('error: missing required argument `version\'');
```
grant: Why does this have a \` character here?
Can we make this contain check have a more readable expected value?

tests/test.ts
```diff
@@ -447,3 +448,112 @@ describe('Test clasp logout function', () => {
...
+  it('should pull error correctly', () => {
+    const result = spawnSync(
+      CLASP, ['pull'], { encoding: 'utf8' },
+    );
+    expect(result.status).to.equal(1);
+    // Should be ERROR.NO_CREDENTIALS
+    // see: https://github.com/google/clasp/issues/278
+    expect(result.stderr).to.contain(ERROR.SCRIPT_ID_DNE);
+  });
+  it('should open error correctly', () => {
```
grant: Don't say "error correctly".
A better description would be: "should error when the script ID DNE"

tests/test.ts
```diff
@@ -447,3 +448,112 @@ describe('Test clasp logout function', () => {
...
+    expect(result.stdout).to.contain(CLASP_USAGE);
+    expect(result.status).to.equal(0);
+  });
+  it('should show help correctly', () => {
+    const result = spawnSync(
+      CLASP, ['-h'], { encoding: 'utf8' },
+    );
+    expect(result.stdout).to.contain(CLASP_USAGE);
+    expect(result.status).to.equal(0);
+  });
```
grant: Prove different descriptions for these cases.
I know it's tedious to write these tests, but they only need to be written correctly once.

tests/test.ts
campionfellin: This is a feature of CommandJS, see https://github.com/tj/commander.js/issues/840

I've updated to skip these tests with a note to see that issue.

grant:
Looks great. Thanks for the contribution @campionfellin!

PR #292, opened 2018-08-16
Add --rootDir option on create command

### LINKED ISSUE #288: clasp create --rootDir ./dist
## Expected Behavior

You should be able to create specify `rootDir`. `rootDir` is specified the local directory in which clasp will store your project files. Also `rootDir` is store in `.clasp.json`.

```
clasp create --rootDir ./dist
```

.clasp.json
```json
{
    "scriptId": "<SCRIPT_ID>",
    "rootDir": "./dist"
}
```

## Actual Behavior

After `clasp create`, You must move pulled files and edit .clasp.json.

```
clasp create --rootDir ./dist
mkdir ./dist
mv appsscritp.json ./dist/
```

fossamagna:
Fixes #288

### DIFF
--- a/src/files.ts
+++ b/src/files.ts
@@ -55,16 +55,16 @@ export function hasProject(): boolean {
 /**
  * Recursively finds all files that are part of the current project, and those that are ignored
  * by .claspignore and calls the passed callback function with the file lists.
- * @param {string | undefined} rootDir The project's root directory
+ * @param {string} rootDir The project's root directory
  * @param {FilesCallBack} callback The callback will be called with the following paramters
  *   error: Error if there's an error, otherwise null
  *   result: string[][], List of two lists of strings, ie. [nonIgnoredFilePaths,ignoredFilePaths]
  *   files?: Array<AppsScriptFile|undefined> Array of AppsScriptFile objects used by clasp push
  */
-export function getProjectFiles(rootDir: string | undefined, callback: FilesCallback): void {
+export function getProjectFiles(rootDir: string = path.join('.', '/'), callback: FilesCallback): void {
   // Read all filenames as a flattened tree
   // Note: filePaths contain relative paths such as "test/bar.ts", "../../src/foo.js"
-  recursive(rootDir || path.join('.', '/'), (err, filePaths) => {
+  recursive(rootDir, (err, filePaths) => {
     if (err) return callback(err, null, null);
     // Filter files that aren't allowed.
     DOTFILE.IGNORE().then((ignorePatterns: string[]) => {
--- a/src/files.ts
+++ b/src/files.ts
@@ -55,13 +55,13 @@ export function hasProject(): boolean {
 /**
  * Recursively finds all files that are part of the current project, and those that are ignored
  * by .claspignore and calls the passed callback function with the file lists.
- * @param {string} rootDir The project's root directory
+ * @param {string | undefined} rootDir The project's root directory
  * @param {FilesCallBack} callback The callback will be called with the following paramters
  *   error: Error if there's an error, otherwise null
  *   result: string[][], List of two lists of strings, ie. [nonIgnoredFilePaths,ignoredFilePaths]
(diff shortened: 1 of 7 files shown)


src/files.ts
```diff
@@ -55,13 +55,13 @@ export function hasProject(): boolean {
 /**
  * Recursively finds all files that are part of the current project, and those that are ignored
  * by .claspignore and calls the passed callback function with the file lists.
- * @param {string} rootDir The project's root directory
+ * @param {string | undefined} rootDir The project's root directory
```
grant: You can say `@param {string?}`

src/utils.ts
```diff
@@ -272,11 +272,16 @@ export async function checkIfOnline() {
...
  * @param  {string} scriptId The script ID
+ * @param  {string} rootDir Local root directory that store your project files
  */
-export async function saveProjectId(scriptId: string): Promise<string> {
-  return DOTFILE.PROJECT().write({ scriptId }); // Save the script id
+export async function saveProject(scriptId: string, rootDir?: string): Promise<ProjectSettings> {
+  const project: ProjectSettings = { scriptId };
+  if (rootDir) {
+    project.rootDir = rootDir;
+  }
```
grant: It's common to write this equivalent statement:
```ts
project.rootDir = project.rootDir || rootDir;
```
instead of
```ts
if (rootDir) {
  project.rootDir = rootDir;
}
```
Can you replace this with the shorter line of code?

src/files.ts
```diff
@@ -55,13 +55,13 @@ export function hasProject(): boolean {
...
  * @param {FilesCallBack} callback The callback will be called with the following paramters
  *   error: Error if there's an error, otherwise null
  *   result: string[][], List of two lists of strings, ie. [nonIgnoredFilePaths,ignoredFilePaths]
  *   files?: Array<AppsScriptFile|undefined> Array of AppsScriptFile objects used by clasp push
  */
-export function getProjectFiles(rootDir: string, callback: FilesCallback): void {
+export function getProjectFiles(rootDir: string | undefined, callback: FilesCallback): void {
```
grant: `rootDir: string?`

fossamagna: ### Result of change

When I changed `rootDir: string?`, I have a error following to:
```
src/commands.ts:506:21 - error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string | null'.
  Type 'undefined' is not assignable to type 'string | null'.

506     getProjectFiles(rootDir, (err, projectFiles) => {
```

### Cause

I consider that the part of the cause is the following part.
https://github.com/google/clasp/blob/6384a2000675158f2e83e198dda1798fd5875abc/src/utils.ts#L45-L49

`rootDir?: string` is same to `rootDir: string | undefined` in property definition. But, it is same to `rootDir: string | null` in parameters definition.

### Solution

I considered that following to as solution:

1. rootDir parameter is changed required. If rootDir is falsy, Pass empty string to rootDir argument when to call getProjectFiles function.
2. No change source code.

grant: Sorry, that my mistake.

Can you try `rootDir?: string`?

If that doesn't work, then we can leave it as-is.

fossamagna: I tried `rootDir?: string`. I have a error following to:
```
> tsc --project tsconfig.json && npm i -g --loglevel=error

src/files.ts:64:51 - error TS1016: A required parameter cannot follow an optional parameter.

64 export function getProjectFiles(rootDir?: string, callback: FilesCallback): void {
                                                     ~~~~~~~~
```

grant: Maybe 
```ts
export function getProjectFiles(rootDir: string = path.join('.', '/'), callback: FilesCallback): void {
```

And then for this line:
```ts
recursive(rootDir || path.join('.', '/'), (err, filePaths) => {
```
we can have
```ts
recursive(rootDir, (err, filePaths) => {
```

fossamagna: I see, it went well.
Thank you.

grant:
Note: I'll probably publish clasp to npm next week.

Issue #296, opened 2018-08-23
Unable to run clasp in a Docker container. 


chris-garrett:
## Expected Behavior

I should be able to run clasp in a docker container. 

## Actual Behavior

Unable to authenticate as localhost service is listening on a random port. 

## Steps to Reproduce the Problem

  1. docker-compose.yml
```
version: '2'

services:

  svc_gsuite:
    image: chrisgarrett/node-dev:8.2.1
    command: npm run up
    volumes:
    - ./src:/work/app
    - ~/.clasprc.json:/home/node/.clasprc.json
```
  2. `up` script entry in package.json:
```
"up": "yarn install && npx clasp login && npm run watch",
```
  3. copy link in and paste into browser. 
  4. redirection fails because port is random. 

## Specifications

  - Node version (`node -v`): 8.2.1
  - Version (`clasp -v`): 1.5.3
  - OS (Mac/Linux/Windows): alpine 3.6

## Options
  1. Use an environment variable to specify port. 
This feels pretty natural as there are many node examples where this is used. I have this working already in my fork 
  1. Add an argument to `login` to specify the port. 

Thoughts?

I will submit a PR for either of these after some feedback.

grant:
Can you be more specific on how you used clasp?

Can you run:
```
clasp login --no-localhost
```

What is the error you see?

See https://github.com/google/clasp#login

chris-garrett:
> Can you be more specific on how you used clasp?

When the container starts up (via docker-compose) the following runs: 
```
yarn install
clasp login
webpack --watch
```
webpack is running `clasp push` when the build is finished (webpack-shell-plugin). 

If I need to authenticate at the `clasp login` step it pauses there until I copy the URL from the log into a browser and once it redirects `clasp login` finishes and it continues on to the webpack step. 

In order for the redirection to work you need to know which port to bind from the container to the host. When it's randomly selected that makes it very difficult to use. 

The error message with 
```
This site can’t be reached
localhost refused to connect.
Search Google for localhost 35087
ERR_CONNECTION_REFUSED
```

This is all happening non-interactive mode which we definitely prefer. 

Hope that makes sense.

grant:
Can you please run

```
clasp login --no-localhost
```

not

```
clasp login
```

And report what you see? You only need to login once.

campionfellin:
Hey @chris-garrett , in your `docker-compose.yaml` file does this:

```
    volumes:
    - ./src:/work/app
[...]

PR #300, opened 2018-08-26
Add the ability to specify the local server's listening port.

### LINKED ISSUE #34: Feature Request - Additions to CLI UI
@grant 
Would you be open to adding some stying to the CLI like Firebase CLI? 
_If so, it would be great to be involved!_

image

I have some suggestions using the same NPM packages that are used in the Firebase CLI. 

Firstly is to add something to make the main clasp text stand out using **figlet.js** (possibly something like the images below but other suggestions are welcome): 
clasp-cli

Second is to add some styling to errors and user input with **chalk** and **Inquirer.js**. An example of this might be to ask the user to set a name for otherwise undefined projects or to get them to select deployment versions. Similar to the Firebase CLI and Webpack CLI.

Thanks 👍

chris-garrett:
Fixes #34
- [ X ] `npm run test` succeeds.
- [ X ] `npm run lint` succeeds.
- [ X ] Appropriate changes to README are included in PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,9 @@ Logs the user in. Saves the client credentials to an rc file.
 - `--no-localhost`: Do not run a local server, manually enter code instead.
 - `--ownkey`: Save .clasprc.json file to current working directory.
 
+##### Environment Variables
+- CLASP_AUTH_PORT - Clasp uses a random port when running the local server. If you need to set the port explicity you can use this environment variable to set it. Ex: CLASP_AUTH_PORT=9000.
+
 ### Logout
 
 Logs out the user by deleting client credentials.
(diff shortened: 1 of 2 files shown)


grant:
I think this addresses #296?

chris-garrett:
Yes sir

grant:
Thanks for the PR. If we'd like to specify the port for logging in, I'd rather have that as a feature flag than a global environment variable. i.e.

```
clasp login --port 8000
```

This would require modifying these files (as well as others):

- https://github.com/google/clasp/blob/9297b98df6c2c11033319b4e440a1fef0ebc84ee/src/index.ts#L73
- https://github.com/google/clasp/blob/95809d546e699ebb554034353029dda11f6c56f6/src/auth.ts#L161

Some more discussion in the issue would be helpful as well as to how we could implement this.

What do you think?

chris-garrett:
Sounds good to me.

grant:
Closing this PR in favor of a new one that follows `login --port` option:
https://github.com/google/clasp/pull/300#issuecomment-416071380

PR #304, opened 2018-08-31
Enable file extension setting

### LINKED ISSUE #293: Custom file extension on `clasp pull`
## Expected Behavior

Right now `clasp pull` defaults Apps Script files to `.js`.

A user should be able to specify which file extension should be used when pulling new files.

Solution: Add a new property to `.clasp.json`:

```
{
  'fileExtension': 'gs'
}
```

Other values could be `'js'` or `'ts'` (or anything else).

## Actual Behavior

All pulled files have `.js`.

sat0yu:
Fixes #293 

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,6 +339,10 @@ Specifies the id of the Google Script project that clasp will target. It is the 
 
 Specifies the **local** directory in which clasp will store your project files. If not specified, clasp will default to the current directory.
 
+### `fileExtension` (optional)
+
+Specifies the file extension for **local** script files in your Apps Script project.
+
 ## Troubleshooting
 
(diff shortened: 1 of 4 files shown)


sat0yu:
I signed it!

grant:
Great job. Thanks!

MassifWelcome:
Thanks for this addition.  It seems to work well for me and resolved a problem because of file extension conflict between doing an using clasp and doing a "git clone".  The issue I still have though is that "clasp clone" still does a rename and changes the .gs to .js .  I'm guessing this is because the .clasp.json file doesn't exist at the point when the "clasp clone" is initiated, so doesn't have the fileExtension doesn't exist at this point.  Is there a switch that can be added to "clasp clone" to stop the renaming behaviour?

PR #312, opened 2018-09-10
Use console.log instead of alert

shouichi:
Because alert doesn't work in Apps Script.


grant:
Thanks!

PR #320, opened 2018-09-12
PR-save-new-project

mainmode1:

### Changes
- **`utils`**
  - `saveProject()` renamed `saveNewProject()` to clarify as not all project settings are from ProjectSettings iface are preserved
- **test**
  - *Test saveNewProject function from utils*: saveProject=>saveNewProject

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -287,7 +287,7 @@ export async function checkIfOnline() {
  * @param  {string} scriptId The script ID
  * @param  {string} rootDir Local root directory that store your project files
  */
-export async function saveProject(scriptId: string, rootDir?: string): Promise<ProjectSettings> {
+export async function saveNewProject(scriptId: string, rootDir?: string): Promise<ProjectSettings> {
   const project: ProjectSettings = { scriptId };
   project.rootDir = project.rootDir || rootDir;
   return DOTFILE.PROJECT().write(project);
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -287,7 +287,7 @@ export async function checkIfOnline() {
  * @param  {string} scriptId The script ID
  * @param  {string} rootDir Local root directory that store your project files
  */
-export async function saveProject(scriptId: string, rootDir?: string): Promise<ProjectSettings> {
+export async function saveNewProject(scriptId: string, rootDir?: string): Promise<ProjectSettings> {
```
grant: The project isn't necessarily 'new'. We're just saving the project settings in the dotfile.

mainmode1: Understood, simply renamed for clarity; only called by `create` & `clone`:
>...not all project settings are from ProjectSettings iface are preserved

Should probably be superseded with originally-named function handling optional parameters so it can be used safely by `pull`, etc. ([#293](https://github.com/google/clasp/issues/293)) so `ProjectSettings` aren't lost.

grant: Currently we only save the project for newly created ones. But it doesn't have to be that way. 

There's nothing in the method that makes old projects not work. Besides, shorter method names are preferred.

PR #328, opened 2018-09-22
Fix broken Travis builds

campionfellin:
I think the credentials issue was that it was not decrypting them to the right place. This fixes that as well as adds documentation for next time the credentials need to be updated.

Also unskips the previously broken tests, and skips a couple that are now broken.

See build: https://travis-ci.org/google/clasp/builds/431920165


grant:
Cool! Makes sense. Let's see if it works.

PR #338, opened 2018-09-27
Fix indentation

abetomo:


grant:
We should also add a lint rule to stop Thisbe from happening again.

PR #340, opened 2018-09-28
Fix pull version option name

### LINKED ISSUE #339: The [--version] option of pull doesn't work
The Apps Script API now allows to retrieve the source code of past versions. With clasp, this was made possible thanks to the [--version] option in the pull command:
`clasp pull --version 23`

But this command returns the current version of clasp (currently returns 1.5.3) instead of pulling the source code of a specified version of an Apps Script project.

abetomo:
In the specification of 'commander', the option with the same name does
not seem to work properly.
It seems that `cmd.version` is always undefined.

Therefore, I changed the pull option name.

Fixes #339 

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ clasp
 - [`clasp logout`](#logout)
 - [`clasp create [scriptTitle] [scriptParentId] [--rootDir]`](#create)
 - [`clasp clone <scriptId>`](#clone)
-- [`clasp pull [--version]`](#pull)
+- [`clasp pull [--versionNumber]`](#pull)
 - [`clasp push [--watch]`](#push)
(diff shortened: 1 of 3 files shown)


grant:
Thanks.
I think `-v` instead of `--projectVersion` would be better. `-v` is pretty common.
Can you change the PR to that?

abetomo:
Thanks for the review.
It changed to `versionNumber`.

RomainVialard:
I see the doc / readme has been updated with `versionNumber`
`clasp pull --versionNumber 23`
I tried version 1.6.0 of clasp but this command throws `error: unknown option --versionNumber`

grant:
@RomainVialard File a bug if there's an issue and I'll fix it.

PR #372, opened 2018-11-02
[WIP] Migrate deploy command to inquirer

takanakahiko:
#57

If you don't input args in `clasp deploy`, we use `inquirer` to ask args.

- [ ] `npm run test` succeeds.
- [x] `npm run lint` succeeds.
- [ ] Appropriate changes to README are included in PR.

Issue #366, opened 2018-10-24
clasp redeploy without version specified


jsarafajr:
## Expected Behavior
It would be good to have the ability to use `clasp redeploy` command without version specified so that clasp would create one for us. Same as `clasp deploy` works - it creates a new version when no version passed as an argument. The use case for this is deploying from CI systems (see below)

## Actual Behavior 
Not sure about other Apps Script projects, but for Gmail add-ons you have to specify a particular Deployment ID which will be available to the users through G Suite marketplace. So every time you're delivering the updates you'll have to update existed deployment instead of creating a new one, which is not that easy to do programmatically, as basically you need:
1. to create a new script version `clasp version ...`
1. somehow retrieve created version number (parse clasp output of the previous command or use googleapis directly)
1. and then you can finally use `clasp redeploy <deploymentId> <created version> <description>`

Probably it would be reasonable to make version parameter optional, e.g.
`clasp redeploy <deploymentId> <description>` - creates new version and updates deployment with that version
`clasp redeploy <deploymentId> <version> <description>` - updates deployment with specified version

grant:
I'll have to make redeploy easier in `clasp@2`.

jsarafajr:
Nice! Good to know @grant 🙂
Is there an ETA for v2?

grant:
Hoping for ~December 14th.
I've been a bit busy giving talks about Apps Script recently. Big talk next week at angularconnect.com

takanakahiko:
I agree with this plan. ( https://github.com/google/clasp/pull/372#issuecomment-435435683 )

Is this plan yet to be decided?
If so, I will create a PR to modify the deploy command.

takanakahiko:
If you do not specify `deployment ID`, we deploy the latest deployment, right?
In that case, I think there is no way to create a new deployment.

grant:
If there's no `deploymentId`, [create a deployment](https://developers.google.com/apps-script/api/reference/rest/v1/projects.deployments/create) with the latest version.

takanakahiko:
@grant 

Umm...
`--description [description]` and `--version [version]` can not be executed on `commander.js`
We seems necessary to change the name of the arguments.

https://github.com/tj/commander.js/issues/516

Example:

`--verrsion [version]` -> `--versionNumber [version]`
`--description [description]` -> `--desc [description]`

PR #376, opened 2018-11-03
Change args in deploy command

takanakahiko:
#366 

This PR follows the design of  https://github.com/google/clasp/pull/372#issuecomment-435435683 .

I am struggling to think of the name of args and I want your idea.
https://github.com/google/clasp/issues/366#issuecomment-435572231

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -517,9 +517,9 @@ export const run = async (functionName: string, cmd: { nondev: boolean }) => {
  * @param cmd.desc {string} The deployment description.
  * @param cmd.deploymentId  {string} The deployment ID to redeploy.
  */
-export const deploy = async (cmd :{
+export const deploy = async (cmd: {
   versionNumber: number,
-  desc: string,
+  description: string,
   deploymentId: string,
 }) => {
   await checkIfOnline();
@@ -527,13 +527,11 @@ export const deploy = async (cmd :{
   const { scriptId } = await getProjectSettings();
   if (!scriptId) return;
   spinner.setSpinnerTitle(LOG.DEPLOYMENT_START(scriptId)).start();
-
-  let versionNumber = cmd.versionNumber;
-  const description = cmd.desc || '';
-  const deploymentId = cmd.deploymentId;
+  let { versionNumber } = cmd;
+  const { description='', deploymentId } = cmd;
 
   // if no version, create a new version
-  if (!versionNumber ){
+  if (!versionNumber){
     const version = await script.projects.versions.create({
       scriptId,
       requestBody: {
@@ -546,10 +544,9 @@ export const deploy = async (cmd :{
     }
     versionNumber = version.data.versionNumber || 0;
     console.log(LOG.VERSION_CREATED(versionNumber));
-    spinner.setSpinnerTitle(LOG.DEPLOYMENT_START(scriptId)).start();
   }
 
-  //spinner.setSpinnerTitle(LOG.DEPLOYMENT_CREATE);
+  spinner.setSpinnerTitle(LOG.DEPLOYMENT_CREATE);
   let deployments;
   if (!deploymentId) { // if no deploymentId, create a new deployment
     deployments = await script.projects.deployments.create({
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -184,15 +184,17 @@ commander
...
+ * @example deploy --versionNumber 4
+ * @example deploy --versionNumber 7 --desc "Updates sidebar logo."
+ * @example deploy --desc "Updates sidebar logo." --deploymentId 123
  */
 commander
-  .command('deploy [version] [description]')
+  .command('deploy')
   .description('Deploy a project')
+  .option('--versionNumber <version>', 'The project version to deploy at.')
+  .option('--desc <description>', 'The deployment description.')
```
grant: `--description`

You can also create shorthand ones if you want:

```
-v, --version
-d, --description
-i, --deploymentId
```

https://www.npmjs.com/package/commander#option-parsing

---

If you create the shorthand, please document it.

src/commands.ts
```diff
@@ -513,50 +513,71 @@ export const run = async (functionName: string, cmd: { nondev: boolean }) => {
...
- * @param version {string} The project version to deploy at.
- * @param description {string} The deployment's description.
+ * @param cmd.versionNumber {string} The project version to deploy at.
+ * @param cmd.desc {string} The deployment description.
+ * @param cmd.deploymentId  {string} The deployment ID to redeploy.
  */
-export const deploy = async (version: number, description = '') => {
+export const deploy = async (cmd :{
+  versionNumber: number,
+  desc: string,
```
grant: `description = ''`,

src/commands.ts
```diff
@@ -513,50 +513,71 @@ export const run = async (functionName: string, cmd: { nondev: boolean }) => {
...
   if (!scriptId) return;
   spinner.setSpinnerTitle(LOG.DEPLOYMENT_START(scriptId)).start();
-  async function createDeployment(versionNumber: number) {
-    spinner.setSpinnerTitle(LOG.DEPLOYMENT_CREATE);
-    const deployments = await script.projects.deployments.create({
+
+  let versionNumber = cmd.versionNumber;
+  const description = cmd.desc || '';
+  const deploymentId = cmd.deploymentId;
+
```
grant: Replace the 3 lines here with:
```ts
let { versionNumber, description, deploymentId } = cmd;
```

src/commands.ts
```diff
@@ -513,50 +513,71 @@ export const run = async (functionName: string, cmd: { nondev: boolean }) => {
...
+  // if no version, create a new version
+  if (!versionNumber ){
```
grant: Nit: space

`if (!versionNumber) {`

Can you run the linter and see if there are any error?

src/commands.ts
takanakahiko: lint error : `Identifier 'desc' is never reassigned; use'const' instead of 'let'.`
I'll Write `const` and `let` separately.

```
let { versionNumber } = cmd;
const { desc='', deploymentId } = cmd;
```

src/commands.ts
takanakahiko: With Typescript, We can not be set Initial value inside type definition...

src/commands.ts
takanakahiko: I didn't get an error when running `npm run lint` at my local environment.
Need to add settings?

src/index.ts
takanakahiko: We can't use `--version` and `-v` in subcommand.
So, I will use `-V` as a temporary measure

takanakahiko:
Sorry,`--description` can be used.
But, we can't use` --version` and `-v` in subcommand.
https://github.com/google/clasp/blob/6d21e0fc9431ecf5996aa117e3abecd098ec9fad/src/index.ts#L355

Therefore, I will use the following

- `-V <version>` `--versionNumber <version>`: The project version to deploy at.
- `-d <description>` `--description <description>`: The deployment description.
- `-i <id>` `--deploymentId <id>`: The deployment ID to redeploy.

grant:
Thanks!

PR #398, opened 2018-11-16
remove `redeploy` command

### LINKED ISSUE #366: clasp redeploy without version specified

takanakahiko:
Fixes #366

Remove `clasp redeploy` command.
This PR follows the design of https://github.com/google/clasp/pull/372#issuecomment-435435683.


grant:
Thanks @takanakahiko!

PR #410, opened 2018-11-20
update ts2gas version

### LINKED ISSUE #408: Translate `export` to nothing or comment
> Please search [existing issues](https://github.com/google/clasp/issues) before creating a new one.
>
> Ideally, after filing the issue, try to fix the issue with a PR!

## Expected Behavior
ignore `export` statements in typescript. 

## Actual Behavior
`export` statements are translated to something like `__export(require("./Utils"));` that leads to error

### General steps to reproduce
1. create ts file with `export` statement (I know it is not supported... that's why I think it should be ignored)
2. push changes to 
### Description of my problem and why I'm doing it
There are several google scripts projects. Some of them use other projects as libraries.
I want to make autocompletion work. But library projects have several files. 
I, of course, cannot do something like this:
``` javascript
import * as UtilsLib from "../utils/Utils" 
import * as UtilsLib from "../utils/Constants" 
```
So I came up with idea to create a `index.ts` file that will export all I need from other files. This file is actually not needed in google script env, but it is needed in development env. 
index.ts
```javascript
export {Goal, BWorks} from './BWorks'
export * from "./Utils"
export * from './Constants'
```
it is translated to something like 
```javascript
var module = module || { exports: exports };
function __export(m) {
    for (var p in m) if (!exports.hasOwnProperty(p)) exports[p] = m[p];
}

var BWorks_1 = require("./BWorks");
exports.Goal = BWorks_1.Goal;
exports.BWorks = BWorks_1.BWorks;
__export(require("./Utils"));
__export(require("./Constants"));
```
This file causes errors due to `require`. If I just comment out the whole file everything is working as expected

I think the right thing will be just to translate `export` to a comment... Or probably create a feature to ignore files... because this file is helpful in local IDE, but not needed at all at scripts.google.com

But probably you can suggested a better way to achieve that I want?

## Specifications

  - Node version (`node -v`): v6.14.4
  - Version (`clasp -v`): 1.6.3
  - OS (Mac/Linux/Windows): mac

pozhidaevak:
Fixes #408  (it's a good idea to open an issue first for discussion)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5553,9 +5553,9 @@
       }
     },
     "ts2gas": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/ts2gas/-/ts2gas-1.2.0.tgz",
-      "integrity": "sha1-GaGk8dG/Iaf3z67Ng+QLKpcIuv0=",
+      "version": "1.3.0",
+      "resolved": "https://registry.npmjs.org/ts2gas/-/ts2gas-1.3.0.tgz",
+      "integrity": "sha1-Y73EuGle7CIbe9fx4KJZMuESVDs=",
       "requires": {
         "typescript": "^2.9.2"
       },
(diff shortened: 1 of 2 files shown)


pozhidaevak:
I signed it!

PR #438, opened 2018-12-06
Fix clasp clone and add more tests

campionfellin:

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -278,24 +278,23 @@ export const clone = async (scriptId: string, versionNumber?: number) => {
       },
     ]);
     scriptId = answers.scriptId;
-  } else {
-    // We have a scriptId or URL
-    // If we passed a URL, extract the scriptId from that. For example:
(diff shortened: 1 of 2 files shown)


src/commands.ts
```diff
@@ -278,24 +278,23 @@ export const clone = async (scriptId: string, versionNumber?: number) => {
...
-    // https://script.google.com/a/DOMAIN/d/1Ng7bNZ1K95wNi2H7IUwZzM68FL6ffxQhyc_ByV42zpS6qAFX8pFsWu2I/edit
-    if (scriptId.length !== 57) {
-      // 57 is the magic number
-      const ids = scriptId.split('/').filter(s => {
-        return s.length === 57;
-      });
-      if (ids.length) {
-        scriptId = ids[0];
-      }
+  }
```
campionfellin: Simply removed this else { } wrapping the logic. Not sure why the diff seems so large.

takanakahiko: By setting it you can hide whitespace changes.

ss

Issue #42, opened 2018-02-13
Add possibility to store login by project


nirazul:
**Feature request**
I have a company project using clasp and a private one. When I'm switching projects, I need to change my clasp login inbetween.
If possible, it'd be nice to store a separate login projectwise instead of a global clasp login.

grant:
I wrote a proposal of how we could store multiple Google account login creds here: https://github.com/google/clasp/issues/12#issuecomment-370672417

Basically, you can set `settingId`s (`[default/global]`/`"work"`/`"personal"`/`"testGoogleAccount"`) that you can tie to projects.
i.e. multilogin support.

grant:
Workaround 1:

You can `clasp login --creds creds.json` with your own credentials from different profiles if you're interested in working with multiple projects right now. This saves the credentials per directory, but it takes some setup.

Workaround 2:

You could also do something like this:
```
clasp login # work
mv ~/.clasprc.json ~/.clasprc-work.json
clasp login # personal
mv ~/.clasprc.json ~/.clasprc-personal.json
```

Switch with:
```
cp ~/.clasprc-personal.json ~/.clasprc.json
```

grant:
## Solution

New flag:

```
clasp login --local
```

Saves `.clasprc.json` locally, but would use global `clasp` credentials.

```
clasp login # global for all projects
clasp login --creds creds.json # your own creds
clasp login --local # local project. Useful for multiple Google accounts.
clasp login --creds creds.json --local # own creds, local
```

See https://github.com/google/clasp/pull/447/files#r240480830

nirazul:
Sounds like a clean solution, thanks!

NoSubstitute:
Is there hope of any progress soon?

kospl:
```
clasp login --local
```
unfortunately, this workaround doesn't work anymore as of `2.3.0` version:

> error: unknown option '--local'

PopGoesTheWza:
Try replacing `@google/clasp` 2.3.0 with `forked-clasp` which is more advanced. : https://github.com/PopGoesTheWza/clasp/blob/forked/docs/config-files.md

```shell
npm uninstall -g @google/clasp
npm install -g forked-clasp
```

PopGoesTheWza:
This should be fixed in https://github.com/google/clasp/releases/tag/v2.3.1

imthenachoman:
[...]

PR #447, opened 2018-12-11
Clean up getOAuthSettings in utils

campionfellin:
The logic for `getOAuthSettings` was originally implemented a few months ago in this commit:

https://github.com/google/clasp/commit/8b65ad7978ab436bbcb461b611d60f6cead1441e

However, this recent commit: 

https://github.com/google/clasp/commit/87a481092a56b8a58d1ceee34e89b5e385b009eb

Changed some things.

My guess is this change was for `clasp run` because the global creds are not sufficient and lead to cryptic error messages, so this was needed to tell users they need local creds for `clasp run`. I think we should do it slightly differently, as we want to keep functionality of searching upward recursively.

I think maybe we can keep the boolean `local` (or change it to `onlyLocalPermitted` (i.e. they are using `clasp run`) and if that's passed as `true` then we don't go up recursively, else, normal logic.

Reduces failed tests from 11 -> 3.  (see: https://travis-ci.org/campionfellin/clasp/jobs/466344222)

My guess is these 3 that are left should be fairly straightforward to fix and are related to other recent commits.

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -187,9 +187,9 @@ export async function authorize(options: {
  * Loads the Apps Script API credentials for the CLI.
  * Required before every API call.
  */
-export async function loadAPICredentials(local = false): Promise<ClaspToken> {
+export async function loadAPICredentials(): Promise<ClaspToken> {
   // Gets the OAuth settings. May be local or global.
-  const rc: ClaspToken = await getOAuthSettings(local);
+  const rc: ClaspToken = await getOAuthSettings();
   await setOauthClientCredentials(rc);
   return rc;
 }
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -187,9 +187,9 @@ export async function authorize(options: {
  * Loads the Apps Script API credentials for the CLI.
  * Required before every API call.
  */
-export async function loadAPICredentials(): Promise<ClaspToken> {
+export async function loadAPICredentials(localCredsOnly = false): Promise<ClaspToken> {
   // Gets the OAuth settings. May be local or global.
-  const rc: ClaspToken = await getOAuthSettings();
+  const rc: ClaspToken = await getOAuthSettings(localCredsOnly);
   await setOauthClientCredentials(rc);
   return rc;
 }
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -598,7 +598,7 @@ export const run = async (functionName: string, cmd: { nondev: boolean }) => {
   async function runFunction(functionName: string) {
     try {
       // Load local credentials.
-      await loadAPICredentials();
+      await loadAPICredentials(true);
       const localScript = await getLocalScript();
       spinner.setSpinnerTitle(`Running function: ${functionName}`).start();
       const res = await localScript.scripts.run({
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -598,7 +598,7 @@ export const run = async (functionName: string, cmd: { nondev: boolean }) => {
   async function runFunction(functionName: string) {
     try {
       // Load local credentials.
-      await loadAPICredentials(true);
+      await loadAPICredentials();
       const localScript = await getLocalScript();
       spinner.setSpinnerTitle(`Running function: ${functionName}`).start();
       const res = await localScript.scripts.run({
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -48,17 +48,25 @@ export const hasOauthClientSettings = (local = false): boolean =>
  * Should be used instead of `DOTFILE.RC?().read()`
  * @returns {Promise<ClaspToken>} A promise to get the rc file as object.
  */
-export function getOAuthSettings(): Promise<ClaspToken> {
-  return DOTFILE.RC_LOCAL()
-    .read()
-    .then((rc: ClaspToken) => rc)
-    .catch((err: any) => {
-      return DOTFILE.RC.read()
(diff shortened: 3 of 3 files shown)


src/commands.ts
```diff
@@ -598,7 +598,7 @@ export const run = async (functionName: string, cmd: { nondev: boolean }) => {
   async function runFunction(functionName: string) {
     try {
       // Load local credentials.
-      await loadAPICredentials(true);
+      await loadAPICredentials();
```
grant: Logically speaking, we cannot use credentials that are global.
We must force local.

src/utils.ts
```diff
@@ -48,12 +48,16 @@ export const hasOauthClientSettings = (local = false): boolean =>
...
-export function getOAuthSettings(local: boolean): Promise<ClaspToken> {
-  const RC = (local) ? DOTFILE.RC_LOCAL() : DOTFILE.RC;
-  return RC.read()
+export function getOAuthSettings(): Promise<ClaspToken> {
+  return DOTFILE.RC_LOCAL()
+    .read()
     .then((rc: ClaspToken) => rc)
     .catch((err: any) => {
-      logError(err, ERROR.NO_CREDENTIALS(local));
+      return DOTFILE.RC.read()
```
grant: I'd rather be explicit about if we're loading local or global credentials.
The previous construct was very nested and hard with nested promises.

If we can make the tests pass with the local/global RC, that would be ideal.

src/auth.ts
```diff
@@ -187,9 +187,9 @@ export async function authorize(options: {
  * Loads the Apps Script API credentials for the CLI.
  * Required before every API call.
  */
-export async function loadAPICredentials(local = false): Promise<ClaspToken> {
```
grant: In some API calls, it makes sense to load only local credentials.
It would be nice to be explicit.

campionfellin:
Note: this relates to the Travis builds because it decrypts the file into the `cwd`, not the `homedir`, and thus when it runs commands that need auth, it cannot find the `.clasprc.json` file.

grant:
Maybe there can be some middle ground where the tests pass, but we're more explicit on which creds we use?

src/auth.ts
campionfellin: Ok, then I think we need an `if/else` where if it is local only, then just check there. If it is false or does not specify, then do it recursively.

campionfellin:
Yes, I think the middle ground is something like:
```
if (local) {
 check local only
} else {
 check recursively
}
```

grant:
Technically both of them are recursive local and global. There's some param `isGlobalCreds` that is set when you `clasp login` now.

I heavily rely on browsing the TS code and don't recall where everything lives, but ideally `clasp run` is the only method that needs to load local creds and every other method can use global creds.

src/utils.ts
```diff
@@ -48,17 +48,25 @@ export const hasOauthClientSettings = (local = false): boolean =>
...
-    });
+export function getOAuthSettings(localCredsOnly: boolean): Promise<ClaspToken> {
+  if (localCredsOnly) {
+    return DOTFILE.RC_LOCAL().read()
+      .then((rc: ClaspToken) => rc)
+      .catch((err: any) => {
+        logError(err, ERROR.NO_CREDENTIALS(localCredsOnly));
+      });
+  } else {
+    return DOTFILE.RC_LOCAL()
```
grant: We should be exclusive.
- local creds are **only** used for `clasp run`
- global creds are used for everything else

I previously thought that local creds should be used for everything once you use them, but talking to some senior eng's at Google convinced me that we should do the above (local only for run).

campionfellin: What about the case where you are managing 2 different `clasp` projects (with different creds)?

```
~/project_1/.clasprc.json
~/project_2/.clasprc.json
```

would work with the recursive strategy.

However, with just

```
~/.clasprc.json
```

Users would have to log back in and out each time.

campionfellin:
Looks like the param is `isLocalCreds`, and if you ctrl+f it, it's used in a bunch of places, so good call out there. That should be looked at before merging this PR.

src/utils.ts
grant: True. That's been an issue. #42
I updated the comment: ~global~ recursive-parent

For the multi-login issue, I think that should be separate from cloud projects.

Something like:

```
clasp login --local
```

This would use global `clasp` credentials but store a token locally.

```
clasp login # global
clasp login --creds creds.json # your own creds
clasp login --local # local
clasp login --creds creds.json --local # own creds, local
```

campionfellin: Ah, I had not seen your comment there yet. I like Workaround 2 as well as the solution proposed here.

Note: does recursive-parent creds include cwd, or is it just parent and above?

grant: local is inclusive:

```
/clasptest/ $ clasp login --local # Log in to Gmail 1
/clasptest/ $ clasp create
/clasptest/ $ echo "function hello() { return 'hi' };" > hello.js
/clasptest/ $ clasp push
```

campionfellin: @grant do you want me to include `--local` in this PR? Or, should I just change the `else {` to check the global creds only?

grant: If we're going to modify login, I'd like to see a change moving towards a `--local` option to solve the variety of issues around multi-login.

grant: Starting with a new PR would be OK.

PR #454, opened 2018-12-13
`saveProject` should not read  if project not found

### LINKED ISSUE #452: `clasp clone` shouldn't rely on `.clasp.json`
## Expected Behavior

`clasp clone 1Q3zCVgK53kEiacR0qvBYFr-A4d720UgZh3cdfDF2oFVJE5SgFiXO0AVZ`

`clasp` clones the project.

## Actual Behavior

```
🐷  clasp clone 1Q3zCVgK53kEiacR0qvBYFr-A4d720UgZh3cdfDF2oFVJE5SgFiXO0AVZ
(node:11110) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open '.clasp.json'
(node:11110) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:11110) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
Cloned 2 files.

No .clasp.json settings found. `create` or `clone` a project first.
```

## Proposed solution:

Looks like the clone function uses `getProjectSettings` to determine the type of file it should clone, see here: https://github.com/google/clasp/blob/master/src/files.ts#L240-L243

I propose that if there are no `.clasp.json` file found, then it defaults to filetype of `js` (let user change to `ts` if they want.

takanakahiko:
Fixes #452 

`Clasp clone` call `saveProject` to setup `.clasp.json` with `ScriptId`
Actualy , `saveProject` will read `.clasp.json` to overwrite it.
So, I fix `saveProject` to don't read `.clasp.json` if project not found.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -242,7 +242,7 @@ export const create = async (cmd: { type: string; title: string; parentId: strin
     saveProject({
       scriptId: createdScriptId,
       rootDir,
-    });
+    }, false);
     if (!manifestExists()) {
       const files = await fetchProject(createdScriptId); // fetches appsscript.json, o.w. `push` breaks
       writeProjectFiles(files, rootDir); // fetches appsscript.json, o.w. `push` breaks
@@ -303,7 +303,7 @@ export const clone = async (scriptId: string, versionNumber?: number) => {
   spinner.setSpinnerTitle(LOG.CLONING);
   saveProject({
     scriptId,
-  });
+  }, false);
   const files = await fetchProject(scriptId, versionNumber);
   await writeProjectFiles(files, '');
 };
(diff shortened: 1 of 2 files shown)


campionfellin:
Could we also explicitly pass `false` to `saveProject` in places like here: 

https://github.com/google/clasp/blob/master/src/commands.ts#L242-L245

(in `clasp create`)  and other uses where you would not ever be appending to an existing `.clasp.json` file.

takanakahiko:
Fixed.

PR #497, opened 2019-01-18
Load tsconfig.json

### LINKED ISSUE #473: Allow manual override of tsconfig used by ts2gas
It appears that clasp uses a hardcoded config object here to transpile its Typescript:
https://github.com/grant/ts2gas/blob/master/index.js#L24

This means I cannot use my own .tsconfig to get read by `clasp push`.

One good use case is concatenating all my files into a single Google Apps Script to avoid execution order issues (I know that can be specified separately) and variables going out of scope depending on their script. 

There are workarounds to this.

## Expected Behavior

Adding the following tsconfig to my project does not strip out comments in the uploaded .gs files:
```
{
	"compilerOptions": {
		"removeComments": true
	},
}

```
## Actual Behavior

Adding the above tsconfig to my project should strip out comments in the uploaded .gs files:

## Specifications

- Node version (`node -v`): v11.5.0
- Version (`clasp -v`): 1.7
- OS (Mac/Linux/Windows): Mac

takanakahiko:
Fixes #473 

When https://github.com/grant/ts2gas/pull/22 is merged and bump up the version of `ts2gas` , I'll corresponds version up within this PR.


PopGoesTheWza:
@takanakahiko please note that tsconfig.json holds a `ts.compilerOptions` object, which is a part of a `ts.transpileOptions` object.
In other words, you should pass to ts2gas something like:
```ts
  const userConf: ts.TranspileOptions = {
    compilerOptions: tsCongAsJson.compilerOptions,  // where tsCongAsJson is the tsconfig file in json format
  };
```

PopGoesTheWza:
@takanakahiko checking `tsconfig` package github, it is based on TypeScript 2.2.x and has not been updated in a while. I am pretty sure the same result can be achieved with just TypeScript 3.2.x and no additional dependencies (I think I commented one issue with tentative code)
IMHO (I don't want to appear as commanding anyone/anything)

takanakahiko:
@PopGoesTheWza Thank you very much. I will fix it as your advised!

grant:
Can you check the tests?
Does `clasp status` still work?

takanakahiko:
Fixed!
Error in test was caused by `JSON.parse` being unable to parse `tsconfig.json` (ex: Trailing commas).

grant:
ty

Issue #501, opened 2019-01-22
Refactor Commands code


campionfellin:
> Please search [existing issues](https://github.com/google/clasp/issues) before creating a new one.
>
> DON'T DELETE THIS! FILL THESE SECTIONS OUT!

## Expected Behavior

The `commands.ts` file is getting quite large and could use some refactoring. This will speed up development and keep a higher quality codebase.

## Actual Behavior

All commands are in the single file.

## Proposed Solution

Move each command under `/src/commands/<command>.ts`

See previous discussion here: https://github.com/google/clasp/issues/134

grant:
Unless @campionfellin refactors all the commands, I don't think this issue will be finished. 🙁

grant:
Re-opened it.

grant:
@campionfellin This suggestion turned out great. Done.

campionfellin:
Great to hear! It's also  interesting to see that many of the commands we're having troubles with, like `clasp logs, push, run` don't have great coverage. Probably by flushing that out we can see what's going wrong.

<img width="469" alt="screen shot 2019-02-17 at 4 58 43 pm" src="https://user-images.githubusercontent.com/11984923/52922330-6f67b080-32d5-11e9-82f3-673cabd831f2.png">

PR #514, opened 2019-01-31
Refactor clasp push

campionfellin:
Pulled `clasp push` out and refactored its internal functions a bit.

Works on #501 


campionfellin:
We also get a lot more granular info on code coverage by separating out each command

screen shot 2019-01-30 at 9 02 37 pm

grant:
Cool!
I'll let @erickoledadevrel review the PR.

---

Can we figure out how to use absolute import paths rather than relative (tsconfig?)?

```ts
} from '/commands/auth'
// Rather than
} from './../auth'
```

campionfellin:
I looked into using absolute paths, but it doesn't look feasible at this time.

We could use absolute path mappings:

```
    "baseUrl": ".",
    "paths": {
      "*": ["*"],
      "src/auth": ["src/auth"],
      "commands/*": ["commands/*"],
    },
  },
  "moduleResolution": "node",
```

as suggested by TS documentation (see https://www.typescriptlang.org/docs/handbook/module-resolution.html).

However, take a look here at the `clone.js` compiled code:

```
var _this = this;
Object.defineProperty(exports, "__esModule", { value: true });
var auth_1 = require("src/auth");
var urls_1 = require("src/urls");
var utils_1 = require("src/utils");
var files_1 = require("src/files");
```

This actually now tries to resolve the modules in the `node_modules` directory.

There's a bunch of discussion here about the issue https://github.com/Microsoft/TypeScript/issues/10866

We could use a 3rd party library, but I'd prefer not to.

campionfellin:
@erickoledadevrel have you had a chance to give this a look? Thanks!

src/commands/push.ts
```diff
@@ -0,0 +1,105 @@
+import {
+  loadAPICredentials,
```
erickoledadevrel: Nitpick: Should this be collapsed into one line, since there is only one import happening?

src/commands/push.ts
```diff
@@ -0,0 +1,105 @@
+import {
```
erickoledadevrel: Given all the imports added to this new file, are there now any unused imports in `commands.ts`? I can't tell just by looking at this PR, but feels like there is a good chance that at least a few of these imports aren't needed by the remaining code.

src/commands/push.ts
```diff
@@ -0,0 +1,105 @@
...
+const inquirer = require('inquirer');
+const prompt = inquirer.prompt;
+inquirer.registerPrompt('autocomplete', require('inquirer-autocomplete-prompt'));
+
+/**
+ * Uploads all files into the script.google.com filesystem.
+ * TODO: Only push the specific files that changed (rather than all files).
+ * @param cmd.watch {boolean} If true, runs `clasp push` when any local file changes. Exit with ^C.
+ */
+export default async (cmd: { watch: boolean, force: boolean }) => {
```
erickoledadevrel: Was there any changes made to the actual code, or was it just copied and pasted as-is? Hard to tell with the diff view.

src/commands/push.ts
campionfellin: It could be. I figured to keep it multi-line to stay with the style of the other imports. I can change to one-line.

src/commands/push.ts
campionfellin: There are many unused imports in `commands.ts` (unrelated).

The only one that was moved over from `commands.ts` to `push.ts` that is now unused in `commands.ts` is `PROJECT_MANIFEST_FILENAME` from `utils`. I can remove it.

src/commands/push.ts
campionfellin: Yes, great question.

The constants `manifestHasChanges` and `confirmManifestUpdate` were extracted to the bottom of the file and comments were added to them.

My thought is that it improves code readability. Additionally, if we wanted in the future, we could export these and test them individually.

PR #515, opened 2019-02-01
Use inlineSourceMaps for better test coverage checks

campionfellin:
General testing improvement... we can now see more accurate coverage numbers and missing lines. 

#### Previously:

screen shot 2019-01-31 at 7 25 43 pm

#### Now

screen shot 2019-01-31 at 7 23 55 pm

(note this is just one test). We can see now it covers the `.ts` files instead of `.js` files.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,6 +9,7 @@
       "es2017"
     ],
     "strict": true,
+    "inlineSourceMap": true,
   },
   "moduleResolution": "node",
   "files": [


grant:
But won't this add source maps to the version published to npm? We probably want this tsconfig option just to be applied when testing.

campionfellin:
I think that should be fine. What would the problem be in publishing source maps to npm?

grant:
Also, thanks for the PRs! I sometimes forget to say thanks.

campionfellin:
Np! It is a good learning opportunity for me!

Issue #391, opened 2018-11-13
[Support Request] clasp run fails even though API deployment exists


j-frost:
## Expected Behavior

`clasp run --dev main` runs function main on the most current API deployment. 

## Actual Behavior

Fails with `Script API executable not published/deployed.`, altough 

```bash
$ clasp deployments   
2 Deployments.
- AKfycbzykaL...qI2EwL1TI @HEAD 
- AKfycbw2xHzsTN8...vUq8CE @2 - oneplatform api meta-version
```

## Steps to Reproduce the Problem

  1. Have project and `clasp push` it
  1. Add version
  1. Deploy version as API executable
  1. Try to run it

## Specifications

  - Node version (`node -v`): v11.1.0
  - Version (`clasp -v`): 1.6.0
  - OS (Mac/Linux/Windows): Linux

grant:
Thanks for the report.
`clasp run` is pretty fragile right now and so I'll check this when I finish `clasp run`.
It would be helpful if you provided more specific repro steps. I.e. what code did your files include?

j-frost:
I've revamped the project into an editor add-on now, so the `main` function is gone. What it did was connect to Sheets and Drive and copy files based on row entries. 

I'm not sure I understand correctly what you mean when you say "what code". It's a customer project, so I can't post the code unfortunately, but I can make a list of which services it uses if that helps you. Is that what you need?

grant:
It’s hard for me to debug without seeing any code or appsscript.json files.

What does your manifest include? Does is have the correct configuration? Deploy via script.google.com to see what you need in your manifest.

grant:
Bump. Can you post some of the `appsscript.json`?

grant:
OK thanks. I'll look into this issue.

obelic2:
Hi, I'm not sure if you have resolved this or not, but I had the same problem of "Script API executable not published/deployed". 

The thing that solved if for me is adding the following in the manifest:
```
"executionApi": {
  "access": "ANYONE"
}
 ```
Before adding this, it only worked on non-dev.

The step is now in the Readme.

<img width="978" alt="screenshot 2018-12-21 13 54 15" src="https://user-images.githubusercontent.com/30458968/50343881-f5ab8480-0528-11e9-90bc-d3ab09776af7.png">

Was run on the latest clasp
Version (clasp -v): 1.7.0

grant:
Please re-open or file a new issue is the instructions are incorrect.
[...]

PR #533, opened 2019-02-15
Improve consistency of information on 'clasp run'

FlorianKempenich:
Improves on #391

I couldn't figure out how to use `clasp run`, even after following all the steps in [/docs/run.md](https://github.com/google/clasp/blob/master/docs/run.md). After some research, I ended up on Issue [#391](https://github.com/google/clasp/issues/391#issuecomment-449383171). 

It turns out the information was already available on the [README.md - Run](https://github.com/google/clasp#run) but wasn't replicated in [/docs/run.md](https://github.com/google/clasp/blob/master/docs/run.md)

I tried to make the process easier to follow by improving the consistency of information across both sources.

### DIFF
--- a/docs/run.md
+++ b/docs/run.md
@@ -15,7 +15,6 @@ To use `clasp run`, you need to complete 4 steps:
     "access": "ANYONE"
   }
   ```
-- _(Optional) Setup Scopes_
 
 #### Setup Instructions
 
--- a/docs/run.md
+++ b/docs/run.md
@@ -4,11 +4,18 @@
 
 ### Prerequisites
 
-To use `clasp run`, you need to complete 3 steps:
+To use `clasp run`, you need to complete 4 steps:
 
 - Set up a **Project ID**.
 - Create an **OAuth Client ID** (Other). Download as `creds.json`.
 - `clasp login --creds creds.json` with this downloaded file.
+- Add the following to `appsscript.json`:
+  ```json
+  "executionApi": {
+    "access": "ANYONE"
+  }
+  ```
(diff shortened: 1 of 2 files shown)


FlorianKempenich:
I signed it!

docs/run.md
```diff
@@ -4,11 +4,18 @@
...
 - Set up a **Project ID**.
 - Create an **OAuth Client ID** (Other). Download as `creds.json`.
 - `clasp login --creds creds.json` with this downloaded file.
+- Add the following to `appsscript.json`:
+  ```json
+  "executionApi": {
+    "access": "ANYONE"
+  }
+  ```
+- _(Optional) Setup Scopes_
```
grant: Can we remove this?
This section is for requirements.

FlorianKempenich: I honestly don't know. I discovered google appscripts exactly 3h before pushing this PR. 
All I know is:
* It was in the main `README.md`
* Without it `clasp run` wouldn't work on my machine

But I **do not** know:
* What it actually does
* If it is required for all project


Are you suggesting to remove it entirely from this document, or to move it to a different section?

I would just like to avoid a situation where newcomers would end up on this page (`/docs/run.md`) before going through the whole `README.md` and wouldn't know how to execute the command, like it happened to me.

grant: Ok.
Please remove this bullet since we don't add information here.

Users can find information under "Run a function that requires scopes"

FlorianKempenich: Removed :)

PR #560, opened 2019-02-25
Refactor and unskip clasp logs test

campionfellin:

### DIFF
--- a/src/commands/logs.ts
+++ b/src/commands/logs.ts
@@ -1,4 +1,3 @@
-import { AxiosResponse } from 'axios';
 import chalk from 'chalk';
 import {logging_v2} from 'googleapis';
 import {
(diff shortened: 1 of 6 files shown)


tests/commands/logs.ts
```diff
@@ -0,0 +1,35 @@
...
+import {
+  cleanup,
+  setupWithoutGCPProject,
+} from '../functions';
+
+describe('Test clasp logs function', () => {
+  before(setupWithoutGCPProject);
+  it('should prompt for logs setup', () => {
+    const result = spawnSync(
+      CLASP, ['logs'], { encoding: 'utf8' },  // --setup is default behaviour
```
erickoledadevrel: If `--setup` is the default, why does this return different output from the 2nd test case below?

tests/commands/logs.ts
```diff
@@ -0,0 +1,35 @@
...
+    expect(result.stdout).to.contain('What is your GCP projectId?');
+  });
+  it('should prompt for logs setup', () => {
+    const result = spawnSync(
+      CLASP, ['logs', '--setup'], { encoding: 'utf8' },
+    );
+    expect(result.status).to.equal(0);
+    expect(result.stdout).to.contain('Open this link:');
+    expect(result.stdout).to.include(`https://script.google.com/d/${SCRIPT_ID}/edit`);
+    expect(result.stdout).to.contain('Go to *Resource > Cloud Platform Project...*');
```
erickoledadevrel: It seems fragile to have a test depend on all of this specific text. Ideally we'd be able to rephrase these strings without breaking tests. Can we just omit these expects?

tests/commands/logs.ts
campionfellin: It looks like it may have been default behavior when these tests were written, but the logic is different now:

https://github.com/google/clasp/blob/190480c9bfbdad4b944fbb0a9c1c17f273d128b8/src/commands/logs.ts#L143-L150

It looks like `--setup` is not default, but if there is no `projectId` in the `.clasp.json` then it exhibits the same behavior as `--setup`.

I think these tests should expect the same result, while testing different situations that lead to it:

https://github.com/google/clasp/blob/190480c9bfbdad4b944fbb0a9c1c17f273d128b8/src/commands/logs.ts#L106-L115

tests/commands/logs.ts
campionfellin: I agree, the source code should probably use strings from here:

https://github.com/google/clasp/blob/190480c9bfbdad4b944fbb0a9c1c17f273d128b8/src/utils.ts#L124

but as seen in comment above, it just uses those strings as-is. I will fix that.

tests/commands/logs.ts
```diff
@@ -12,24 +12,29 @@ import {
   setupWithoutGCPProject,
 } from '../functions';
 
+import {
+  LOG,
+} from '../../src/utils';
+
 describe('Test clasp logs function', () => {
```
erickoledadevrel: Change the description of this test, since it's not actually testing the logging function, it's also confirming the setup.

tests/commands/logs.ts
erickoledadevrel: I just don't see why it's necessary to test each of those lines at all. I think the test is trying to confirm that the command will print the setup text, but you know that's true after you check the first line.

campionfellin: Great point, I will change it to just expect the prompt "What is your GCP projectId?"

tests/commands/logs.ts
campionfellin: Ok, I will change the description and write additional tests for the rest of the logging function.

src/commands/logs.ts
```diff
@@ -34,111 +33,7 @@ inquirer.registerPrompt('autocomplete', require('inquirer-autocomplete-prompt'))
  */
 export default async (cmd: { json: boolean; open: boolean; setup: boolean; watch: boolean }) => {
   await checkIfOnline();
-  /**
```
erickoledadevrel: Did you just reorder this code, or did you also make edits to it? Hard to tell with the GitHub diff view.

campionfellin: Mostly just extracted these inner functions to outside the main one, small edits noted below.

src/commands/logs.ts
```diff
@@ -155,78 +49,188 @@ export default async (cmd: { json: boolean; open: boolean; setup: boolean; watch
...
+    fetchAndPrintLogs(cmd.json, projectId);
+  }
+};
+
+/**
+ * Prints log entries
+ * @param entries {any[]} StackDriver log entries.
+ */
+export function printLogs(entries: logging_v2.Schema$LogEntry[] = [],
+                   formatJson: boolean) {
```
campionfellin: Now additionally takes `formatJson` parameter (which corresponds to `cmd.json`) because it cannot access the `cmd` object from outside the main function.

src/commands/logs.ts
```diff
@@ -155,78 +49,188 @@ export default async (cmd: { json: boolean; open: boolean; setup: boolean; watch
...
+      protoPayload = {},
+      jsonPayload = null,
+      insertId = '',
+    } = entries[i];
+    if (!resource || !resource.labels) return;
+    let functionName = resource.labels.function_name;
+    functionName = functionName ? padEnd(functionName, 15) : ERROR.NO_FUNCTION_NAME;
+    // tslint:disable-next-line:no-any
+    let payloadData: any = '';
+    if (formatJson) {
```
campionfellin: `cmd.json` --> `formatJson`

src/commands/logs.ts
```diff
@@ -155,78 +49,188 @@ export default async (cmd: { json: boolean; open: boolean; setup: boolean; watch
...
+    spinner.stop(true);
+  });
+  return promise;
+}
+
+/**
+ * Fetches the logs and prints the to the user.
+ * @param startDate {Date?} Get logs from this date to now.
+ */
+export async function fetchAndPrintLogs(formatJson: boolean, projectId?: string, startDate?: Date) {
```
campionfellin: Now takes `formatJson`, and also now takes optional `projectId` (which was previously accessible in main function)

src/commands/logs.ts
```diff
@@ -155,78 +49,188 @@ export default async (cmd: { json: boolean; open: boolean; setup: boolean; watch
...
+          case 403:
+            logError(
+              null,
+              oauthSettings.isLocalCreds ? ERROR.PERMISSION_DENIED_LOCAL : ERROR.PERMISSION_DENIED,
+            );
+          default:
+            logError(null, `(${logs.status}) Error: ${logs.statusText}`);
+        }
+      } else {
+        printLogs(logs.data.entries, formatJson);
```
campionfellin: `cmd.json` --> `formatJson`

erickoledadevrel:
@grant - Can you give this a quick once over, to make sure I didn't miss anything?

tests/test.ts
```diff
@@ -638,7 +611,7 @@ describe.skip('Test clasp run function', () => {
     const result = spawnSync(
       CLASP, ['run', 'myFunction'], { encoding: 'utf8' },
     );
-    expect(result.stdout).to.contain('What is your GCP projectId?');
+    expect(result.stdout).to.contain(`${LOG.ASK_PROJECT_ID}`);
```
grant:
```ts
expect(result.stdout).to.contain(LOG.ASK_PROJECT_ID);
```

campionfellin: Will fix this then merge once build passes.

PR #565, opened 2019-03-07
Refactor and rewrite clasp run and manifest tests

campionfellin:
Refactors the `clasp run` tests, as well as a number of functions within `run.ts` and `manifest.ts`. Comments inline.

Increases overall coverage from:

```
-----------------|----------|----------|----------|----------|-------------------|
File             |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-----------------|----------|----------|----------|----------|-------------------|
All files        |    73.85 |    49.83 |    77.17 |    75.66 |                   |
 src             |    67.84 |    54.69 |    62.69 |    70.31 |                   |
```
to
```
-----------------|----------|----------|----------|----------|-------------------|
File             |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-----------------|----------|----------|----------|----------|-------------------|
All files        |    76.69 |    52.55 |    81.22 |    78.38 |                   |
 src             |    71.22 |    59.23 |    68.61 |    73.78 |                   |
```

Specifically, `clasp run` coverage from:

`  run.ts         |    12.33 |        0 |        0 |    13.11 |`

to

`run.ts         |    37.68 |    11.54 |       50 |    35.09 |`

and manifest from:

`manifest.ts    |    62.96 |    54.17 |    66.67 |    65.08 |`

to

`manifest.ts    |    67.05 |    60.71 |    72.73 |    69.12 |`

and utils from:

`utils.ts       |    66.67 |    53.93 |    61.67 |    65.71 |`

to

`utils.ts       |     67.5 |    56.18 |     62.3 |    66.67 |`

### DIFF
--- a/src/commands/run.ts
+++ b/src/commands/run.ts
@@ -10,13 +10,14 @@ import {
 } from '../auth';
 import {
   addScopeToManifest,
-  isValidManifest,
+  isValidRunManifest,
 } from '../manifest';
 import { URL } from '../urls';
 import {
   ERROR,
   checkIfOnline,
   getProjectSettings,
+  getValidJSON,
   logError,
   spinner,
 } from '../utils';
@@ -24,32 +25,20 @@ import {
  * Executes an Apps Script function. Requires clasp login --creds.
  * @param functionName {string} The function name within the Apps Script project.
  * @param cmd.nondev {boolean} If we want to run the last deployed version vs the latest code.
+ * @param cmd.params {string} JSON string of parameters to be input to function.
  * @see https://developers.google.com/apps-script/api/how-tos/execute
  * @requires `clasp login --creds` to be run beforehand.
  */
 export default async (functionName: string, cmd: { nondev: boolean; params: string }) => {
-  function IsValidJSONString(str: string) {
-    try {
-      JSON.parse(str);
-    } catch (error) {
-      throw new Error('Error: Input params not Valid JSON string. Please fix and try again');
-    }
-    return true;
-  }
 
   await checkIfOnline();
   await loadAPICredentials();
   const { scriptId } = await getProjectSettings(true);
   const devMode = !cmd.nondev; // defaults to true
   const { params: paramString = '[]' } = cmd;
-  IsValidJSONString(paramString);
-  const params = JSON.parse(paramString);
-  // Ensures the manifest is correct for running a function.
-  // The manifest must include:
-  // "executionApi": {
-  //   "access": "MYSELF"
-  // }
-  await isValidManifest();
+  const params = getValidJSON(paramString);
+
+  await isValidRunManifest();
 
   // TODO COMMENT THIS. This uses a method that gives a HTML 404.
   // await enableExecutionAPI();
@@ -59,6 +48,7 @@ export default async (functionName: string, cmd: { nondev: boolean; params: stri
   // - Ensure the execution API is enabled.
   // - Ensure we can run functions that were developed locally but not pushed.
   if (devMode) {
+    console.log('Running in dev mode.');
     // TODO enable this once we can properly await pushFiles
     // await pushFiles(true);
   }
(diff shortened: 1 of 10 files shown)


src/commands/run.ts
```diff
@@ -10,46 +10,35 @@ import {
...
 /**
  * Executes an Apps Script function. Requires clasp login --creds.
  * @param functionName {string} The function name within the Apps Script project.
  * @param cmd.nondev {boolean} If we want to run the last deployed version vs the latest code.
+ * @param cmd.params {string} JSON string of parameters to be input to function.
  * @see https://developers.google.com/apps-script/api/how-tos/execute
  * @requires `clasp login --creds` to be run beforehand.
  */
 export default async (functionName: string, cmd: { nondev: boolean; params: string }) => {
-  function IsValidJSONString(str: string) {
```
campionfellin: Moved this to `src/utils.ts` and changed it to `getValidJSON()`.

src/commands/run.ts
```diff
@@ -10,46 +10,35 @@ import {
...
-  const params = JSON.parse(paramString);
-  // Ensures the manifest is correct for running a function.
-  // The manifest must include:
-  // "executionApi": {
-  //   "access": "MYSELF"
-  // }
-  await isValidManifest();
+  const params = getValidJSON(paramString);
+
+  await isValidRunManifest();
```
campionfellin: `isValidManifest` function simply tested that it was valid JSON. This makes sure it includes 
```
"executionApi": {	
    "access": "MYSELF"	
  }
```

as noted in the comment.

src/commands/run.ts
```diff
@@ -59,113 +48,115 @@ export default async (functionName: string, cmd: { nondev: boolean; params: stri
...
-        }
-      } else if (data.error && data.error.details) {
-        // @see https://developers.google.com/apps-script/api/reference/rest/v1/scripts/run#Status
-        console.error(
-          `${chalk.red('Exception:')}`,
-          data.error.details[0].errorType,
-          data.error.details[0].errorMessage,
-          data.error.details[0].scriptStackTraceElements || [],
-        );
+  await runFunction(functionName, params, scriptId, devMode);
```
campionfellin: Moved `runFunction` out of the main function into its own. No changes here.

src/manifest.ts
```diff
@@ -46,22 +47,40 @@ export async function writeManifest(manifest: Manifest) {
...
+
+/**
+ * Reads manifest file from project root dir.
  * The manifest is valid if it:
  * - It exists in the project root.
  * - Is valid JSON.
  */
-export async function isValidManifest(): Promise<boolean> {
+// tslint:disable-next-line:no-any
+export async function getManifest(): Promise<any> {
```
campionfellin: `isValidManifest` previously just checked if it was a valid JSON file, which is what `getValidJSON` does anyway. This returns a manifest if its valid.

src/commands/run.ts
erickoledadevrel: Should the other access levels be allowed to? If you've deployed it to run for your entire domain, then that should be good enough for clasp run.

src/manifest.ts
```diff
@@ -46,22 +47,40 @@ export async function writeManifest(manifest: Manifest) {
...
+
+/**
+ * Ensures the manifest is correct for running a function.
+ * The manifest must include:
+ * "executionApi": {
+ *   "access": "MYSELF"
+ * }
+ */
+export async function isValidRunManifest(): Promise<boolean> {
+  if (isValidManifest) {
```
erickoledadevrel: Shouldn't you be running this function? Doesn't this current like just check if the function is defined?

src/commands/run.ts
campionfellin: Definitely. Is there documentation on the other access levels? Or should It just be a blanket of anything containing `executionApi` and `access`, regardless of value of `access`?

src/manifest.ts
campionfellin: Yes, great catch, should be `if (isValidManifest()) {`. I'll look at adding another test for this as well.

erickoledadevrel: And that function returns a promise, so I think you'll also need to await the promise to get the actual boolean result.

src/commands/run.ts
erickoledadevrel: The values are documented here:

https://developers.google.com/apps-script/concepts/manifests

All of the values would seem to work for the `clasp run` case, so I don't think we need to check which access is set.

tests/constants.ts
```diff
@@ -9,12 +9,27 @@ import { OAuth2ClientOptions } from 'google-auth-library';
...
+const APPSSCRIPT_JSON_WITH_RUN = JSON.stringify({
+  timeZone: 'America/Los_Angeles',
+  dependencies: {},
+  exceptionLogging: 'STACKDRIVER',
+  executionApi: {
+    access: 'MYSELF',
+  },
+});
+
+export const TEST_APPSSCRIPT_JSON = {
```
grant: Please remove this const and export the other constants.

Issue #512, opened 2019-01-30
Incorrect display name of available "Advanced Service" APIs


tehhowch:
## Expected Behavior
When using `clasp apis list`, the displayed `<api name>` is what can be passed in `clasp apis enable/disable <api name>`

## Actual Behavior
Services with names that have "-" (e.g. `bigquery-json`) or specialized "Advanced Services" bindings (`AdminDirectory`, `AdminReports`) are not properly reported by `apis list`:
![image](https://user-images.githubusercontent.com/20871346/51994620-cd436e00-2476-11e9-86d0-75ee55cdec16.png)

If attempting to then enable the API by displayed name, e.g. `clasp apis enable bigquery`, HTTP 403 "The caller does not have permission" is raised, and clasp then states:
>API bigquery doesn't exist. Try 'clasp apis enable sheets'.

## Steps to Reproduce the Problem

1. `clasp apis list`
   - shows `admin` twice (for `AdminDirectory` & `AdminReports`)
   - shows `bigquery`
2. `clasp apis enable bigquery`
3. Receive large error trace re: HTTP 403, and message "API bigquery doesn't exist. Try 'clasp apis enable sheets'."

## Specifications

- Node version (`node -v`): 8.11.2
- Version (`clasp -v`): 2.0.1
- OS (Mac/Linux/Windows): Win 10

erickoledadevrel:
Thanks for issue report. There seems to be two distinct issues here caused by some faulty assumptions:

1. That the service ID used in Apps Script will match the prefix of the serviceName used in the serviceusage API. While almost always true, for the BigQuery API the Apps Script service ID is `bigquery` but the serviceName is `bigquery-json.googleapis.com`. See assumption in code [here](https://github.com/google/clasp/blob/f00bbbf019befd79e5d82790c5647bdfb82c534c/src/apiutils.ts#L81).
1. That every service ID maps to only one advanced service. While almost always true, the `admin` service ID maps to both the `AdminDirectory` and `AdminReports` advanced services, distinguished by the version. See assumption in code [here](https://github.com/google/clasp/blob/f00bbbf019befd79e5d82790c5647bdfb82c534c/src/manifest.ts#L135).

[...]

PR #585, opened 2019-03-26
Add People API support

### LINKED ISSUE #559: Missing clasp support for People API
## Expected Behavior
People API can be enabled via the web UI. The API should be on when adding it to `appsscript.json` like this
`...
"dependencies": {
    "enabledAdvancedServices": [      {
        "userSymbol": "People",
        "serviceId": "people",
        "version": "v1"
      }
    ]
  }
...`

## Actual Behavior
After `clasp push` PEOPLE API is switched off. Looks like this happens because the `people` constant is not added to the `PUBLIC_ADVANCED_SERVICES` (https://github.com/google/clasp/blob/f00bbbf019befd79e5d82790c5647bdfb82c534c/src/apis.ts#L38)

1. In appscript.json add a snippet from above 
2. `clasp push`
3. In project's web UI -> Resources -> Advanced Google services... 
4. People API is OFF

## Specifications

- Node version (`node -v`): v10.14.2
- Version (`clasp -v`): 1.6.3
- OS (Mac/Linux/Windows): Windows

tawAsh1:
Fixes #559  (it's a good idea to open an issue first for discussion)

Add People API description to AdvancedService[]

### DIFF
--- a/src/apis.ts
+++ b/src/apis.ts
@@ -171,6 +171,11 @@ export const PUBLIC_ADVANCED_SERVICES: AdvancedService[] = [
     serviceId: 'analytics',
     version: 'v3',
   },
+  {
+    userSymbol: 'People',
+    serviceId: 'peopleapi',
+    version: 'v1',
+  },
 ];
 
 export const SCRIPT_ID_LENGTH = 57;


src/apis.ts
```diff
@@ -171,6 +171,11 @@ export const PUBLIC_ADVANCED_SERVICES: AdvancedService[] = [
     serviceId: 'analytics',
     version: 'v3',
   },
+  {
+    userSymbol: 'People',
+    serviceId: 'peopleapi',
```
erickoledadevrel: Are you sure "peopleapi" is correct? Looking at the Service Usage API, it seems like it should just be "people".

tawAsh1: You are right. When I use Service Usage API, it says serviceId is `people`.

However, If I clone the script with PeopleAPI enabled on the GUI, appscript.json says `serviceId:" peopleapi "`.

erickoledadevrel: Ah, you are right. This seems like another case of issue #512. It was assumed that the `serviceId` in  `appsscript.json` would be the same as the prefix used by the Service Usage API, but that isn't always the case. Until that issue is fixed we won't be able to support the People API fully.

grant:
Do you want to close this PR until #512 is fixed?

Issue #587, opened 2019-03-29
.claspignore not working at all


Edge-Developer:
## Expected Behavior

The text below written within `.claspignore` is supposed to ignore everything else and select sample.js for push

```
**/**
!sample.js
```

## Actual Behavior
But Alas, after running `clasp status` on the terminal

it did not ignore the rest:
<figure>
<img width="344" alt="'clasp status' on terminal" src="https://user-images.githubusercontent.com/22748496/55210437-b6579880-51e7-11e9-93d6-0e610babd892.png">
  <figcaption>"clasp status" on terminal</figcaption>
</figure>
<br><br>
Still did not ignore when i pushed
<figure>
<img width="318" alt="Screen Shot 2019-03-29 at 6 01 52 AM" src="https://user-images.githubusercontent.com/22748496/55210556-43025680-51e8-11e9-8002-7fc2e38df9fc.png">
  <figcaption>"clasp push" on terminal</figcaption>
</figure>

## Steps to Reproduce the Problem

Already given above

## Specifications
- Node version: 10.15.3
- Clasp Version: 2.1.0
- OS: Mac (Mojave)

grant:
Thanks for the report. Probably related to #58.

PopGoesTheWza:
This also needs to be validated after PR #620

PopGoesTheWza:
@Edge-Developer  can you try with clasp 2.2.0 if the issue still occurs?

PR #590, opened 2019-03-29
refactor: remove extra fs tree read on getProjectFiles

grant:
So, clasp read the fs tree 2x before `clasp push` or `clasp status`...
This removes the extra unused read.

Simplify pushing. Part of #587.

### DIFF
--- a/src/files.ts
+++ b/src/files.ts
@@ -88,106 +88,103 @@ export async function getProjectFiles(rootDir: string = path.join('.', '/'), cal
     const file2path: Array<{ path: string; file: AppsScriptFile }> = [];  // used by `filePushOrder`
     let ignoredFilePaths: string[] = [];
     ignoredFilePaths = ignoredFilePaths.concat(ignorePatterns);
-    // Match the files with ignored glob pattern
-    readMultipleFiles(filePaths, 'utf8', (err: string, contents: string[]) => {
-      if (err) return callback(new Error(err), null, null);
-      // Check if there are files that will conflict if renamed .gs to .js.
-      // When pushing to Apps Script, these files will overwrite each other.
-      filePaths.map((name: string) => {
-        const fileNameWithoutExt = name.slice(0, -path.extname(name).length);
-        if (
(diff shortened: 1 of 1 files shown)


src/files.ts
```diff
@@ -88,106 +88,103 @@ export async function getProjectFiles(rootDir: string = path.join('.', '/'), cal
...
-            // only print error once (for .gs)
-            logError(null, ERROR.CONFLICTING_FILE_EXTENSION(fileNameWithoutExt));
-          }
+
+    // Check if there are files that will conflict if renamed .gs to .js.
+    // When pushing to Apps Script, these files will overwrite each other.
+    filePaths.map((name: string) => {
+      const fileNameWithoutExt = name.slice(0, -path.extname(name).length);
+      if (
+        filePaths.indexOf(fileNameWithoutExt + '.js') !== -1 &&
```
erickoledadevrel: Optional: Use `includes()` here:

```js
filePaths.includes(fileNameWithoutExt + '.js')
```

src/files.ts
```diff
@@ -88,106 +88,103 @@ export async function getProjectFiles(rootDir: string = path.join('.', '/'), cal
...
+        abortPush = true;
+        if (path.extname(name) === '.gs') {
+          // only print error once (for .gs)
+          logError(null, ERROR.CONFLICTING_FILE_EXTENSION(fileNameWithoutExt));
         }
-      });
-      if (abortPush) return callback(new Error(), null, null);
+      }
+    });
+    if (abortPush) return callback(new Error(), null, null);
```
erickoledadevrel: Seems odd the `Error` has no message string.

grant: This is because clasp's error handler doesn't use `Error` descriptions.

This is one of the oldest methods in `clasp and has a poor interface. It really needs to be refactored:

`callback(new Error(), null, null);` - makes no sense out of context

src/files.ts
grant: Good refactor. Not going to do this in this PR, but we should remove `indexOf` everywhere:

https://github.com/google/clasp/search?q=indexOf&unscoped_q=indexOf

Issue #593, opened 2019-03-30
Travis Tests are Slow


grant:
## Expected Behavior

`clasp` Travis tests should run faster.

## Actual Behavior

Travis is really slow. ~10 minutes per build.

## Possible Solutions

Reduce/combine simple tests. Example `clasp * --help` takes 2 minutes:

```sh
Test --help for each function
    ✓ should run --help (6492ms)
    ✓ should logs --help (6508ms)
    ✓ should login --help (6605ms)
    ✓ should logout --help (6712ms)
    ✓ should create --help (6818ms)
    ✓ should clone --help (6941ms)
    ✓ should pull --help (6691ms)
    ✓ should push --help (6681ms)
    ✓ should status --help (6688ms)
    ✓ should open --help (6715ms)
    ✓ should deployments --help (6711ms)
    ✓ should undeploy --help (6652ms)
    ✓ should versions --help (6779ms)
    ✓ should version --help (6586ms)
    ✓ should list --help (6684ms)
    ✓ should apis --help (6687ms)
    ✓ should help --help (6832ms)
```

- Maybe `spawnSync` is slow and we should just `spawnSync` once instead of dozens of times?
  - https://github.com/google/clasp/blob/147b23c087d42be2f365ea3c24cea90a38dc7178/tests/test.ts#L36
- Maybe we should spawn child processes for some tests?

grant:
It looks like we should speed up the actual commands.

Commander takes ~0.2s
clasp * -h takes ~2s

Need to investigate this.

```sh
time (clasp push -h)

real	0m2.196s
user	0m1.326s
sys	0m0.429s
```

campionfellin:
Additionally, we can look into using `mocha-parallel-tests` (https://github.com/mocha-parallel/mocha-parallel-tests). We would need to finish refactoring, but it looks promising.

PopGoesTheWza:
There is a big difference in Travis time to process depending on the time of the day

PR #594, opened 2019-03-30
refactor: Move clasp * --help test to one test with many assertions

grant:
Move clasp * --help test to one test with many assertions.

Helps with #593.

---

This actually isn't helping yet:

```
Test --help for each function
    ✓ should print help for clasp * --help (85076ms)
```

### DIFF
--- a/tests/test.ts
+++ b/tests/test.ts
@@ -40,23 +40,25 @@ describe('Test --help for each function', () => {
     expect(result.status).to.equal(0);
     expect(result.stdout).to.include(expected);
   };
-  it('should run --help', () => expectHelp('run', 'Run a function in your Apps Scripts project'));
-  it('should logs --help', () => expectHelp('logs', 'Shows the StackDriver logs'));
-  it('should login --help', () => expectHelp('login', 'Log in to script.google.com'));
-  it('should logout --help', () => expectHelp('logout', 'Log out'));
-  it('should create --help', () => expectHelp('create', 'Create a script'));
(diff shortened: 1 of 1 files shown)


campionfellin:
LGTM, but I'm not sure it would be much faster. Looks like it's timing out on some of the builds.

Can you put in test-level timeout: https://mochajs.org/#test-level

Setting `this.timeout(0)` should just disable the timeout.

Also perhaps, looking forward, we can skip these tests and skip the coverage on the`--help` functionality?

erickoledadevrel:
Even after the refactor it's still running `clasp <command> --help` once for each `expectHelp` call, and that's probably where the time is coming from.

More generally, I'm not a fan of tests that are checking for exact strings to be returned. I think you should be free to completely change the help text without it breaking tests.

campionfellin:
I disagree. I think that changing the help texts should break tests, making us more conscious of making those changes.

That being said, I'm fine with skipping all of these tests for now and taking the coverage hit, as it will help with velocity.

erickoledadevrel:
IMO the tests should be confirming that the "contract" is met; that the program does what it claims it will do. The contact for the `--help` is that we will show some relevant help text for the command, not that we'll show a very specific piece of text. I think the point of these tests is to ensure that every command has help text, not to prevent the help text from changing. When testing a command line tool like this it may be hard to test one without testing the other, but I think we should strive to separate those when possible.

grant:
Closing:
- This PR isn't helping
- The tests need to be more useful to be refactored as Eric mentions.

PR #595, opened 2019-04-01
Reduce number of test jobs from 8 to 3

grant:
- Reduce the number of Travis jobs to prevent API resources from being exhausted.
- Keep core job types. Removed job types don't uniquely fail.

@erickoledadevrel What do you think of this mitigation of Travis failures?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 sudo: required
 os:
 - linux
-- osx
+# - osx
 language: node_js
 node_js:
-- node
+# - node
 - '10'
 - '8'
 - '6'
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,10 +1,10 @@
 sudo: required
 os:
 - linux
-- osx
+# - osx
 language: node_js
 node_js:
-- node
+# - node
```
campionfellin: I would include this (it uses latest version) or explicitly put 11.

Additionally, Node 6 EOL is at the end of the month, so we can go ahead and remove it.

grant: Done.

Issue #604, opened 2019-04-29
(Minor) npm package `readline` wrongly added in place of node `readline` module


PopGoesTheWza:
## Steps to Reproduce the Problem

1. check clasp's packages.json for `readline` depenecy
1. check [`readline` npm page](https://www.npmjs.com/package/readline) for renaming warning
1. check [`linebyline` npm page](https://www.npmjs.com/package/linebyline) to verify same author/same api

## Specifications

- Node version (`node -v`): n.a.
- Version (`clasp -v`): @latest
- OS (Mac/Linux/Windows): n.a.

PopGoesTheWza:
After closer inspection, the npm package `readline` is never effectively `require`d and the standard node module `readline` is in fact what is being used. Thus, droping the npm package is the answer.

grant:
Fixed with #606.

PR #606, opened 2019-05-05
Package dependencies update and linting/typing improvement

### LINKED ISSUE #603: (Minor) deprecated `opn` packages should be replace with its successor `open`

1. check clasp's packages.json for `opn` depenecy
1. check [`opn` npm page](https://www.npmjs.com/package/opn) for deprecation warning
1. check [`open` npm page](https://www.npmjs.com/package/open) to verify same author/same api

## Specifications

- Node version (`node -v`): n.a.
- Version (`clasp -v`): @latest
- OS (Mac/Linux/Windows): n.a.

PopGoesTheWza:
Fixes #603 #604 


PopGoesTheWza:
updating `nyc` package, partly addressing #607

PopGoesTheWza:
@grant can someone review this PR please?

src/utils.ts
```diff
@@ -400,7 +399,7 @@ function getScriptTypeName(type: string) {
  * Handles error of each command.
  */
 // tslint:disable-next-line:no-any
-export function handleError(command: (...args: any[]) => Promise<void>) {
+export function handleError(command: (...args: any[]) => Promise<any>) {
```
grant: `void` is better than `any`. I think this returns `void` and should be kept this way.

grant:
Thanks for waiting after Google I/O last week.
Please use *branches* rather than `PopGoesTheWza:master` for future PRs to prevent so many commits.

Will merge after you address the comment.

src/utils.ts
PopGoesTheWza: @grant There are three instances of `handleError` usage which fails when `Promise<void>` is expected.

```ts
  .action(handleError(openCmd));
```

```ts
  .action(handleError(logs));
```

```ts
  .action(handleError(apis));
```

Those three functions return `Promise<ChildProcess | undefined>`

To fix those linting errors, we can either redefine `handleError` to accept either `(...args: any[]) => Promise<any>` or `(...args: any[]) => Promise<unknown>`

Else `openCmd`, 'logs' and `apis` require to be adjusted somehow. Since I am not familiar with these parts of Clasp, I did not invest too much in the latest option.

Please advise.

PopGoesTheWza:
> Please use _branches_ rather than `PopGoesTheWza:master` for future PRs to prevent so many commits.

I will look into how branches work and proceed so in the future.

README.md
```diff
@@ -509,7 +515,7 @@ This requires using the environment variables `HTTP_PROXY` / `HTTPS_PROXY`.
 
 Using clasp for your project? Add a README badge to show it off: [![clasp](https://img.shields.io/badge/built%20with-clasp-4285f4.svg)](https://github.com/google/clasp)
 
-```
+```text
```
grant: \`\`\`md

src/commands/openCmd.ts
```diff
@@ -87,13 +87,18 @@ export default async (scriptId: string, cmd: {
...
-  }]) as {deploymentId: string};
+  }]);
   const deployment = await script.projects.deployments.get({deploymentId: answers.deploymentId});
   console.log(LOG.OPEN_WEBAPP(answers.deploymentId));
-  open(getWebApplicationURL(deployment.data), { wait: false });
+  const target = getWebApplicationURL(deployment.data);
+  if (target) {
+    open(target, { wait: false });
+  } else {
+    // TODO: handle undefined
```
grant: Return something:
```ts
return logError(null, `Could not open deployment: ${deployment}`);
```

src/files.ts
```diff
@@ -9,8 +9,10 @@ import { DOT, DOTFILE } from './dotfile';
 import { ERROR, LOG, checkIfOnline, getAPIFileType, getProjectSettings, logError, spinner } from './utils';
 
 const ts2gas = require('ts2gas');
+// TODO: include ts2gas .d.ts in package
```
grant: Can we do this now? What needs to change?

PopGoesTheWza: @grant currently grant/ts2gas has:

[x] tsconfig.json with `declaration` set to `true`
[x] package.json with `types` set to `./src/index.d.ts`
[ ] `.d.ts` (and optionaly `.ts` included in the published package

Last point is because of `.npmignore` which excludes any `*.ts`

If you want, I could issue a new PR (from a `branch`) with the `.npmignore` change, and possibly the additionnal changes:

- `src` folder contains only the source `.ts`
- a new `lib` folder which will contains `.js` and `.d.ts``
- `package.json` attributes `main` and `types` to properly point to the `lib` folder

grant: See https://github.com/grant/ts2gas/issues/39

Let's fix it and publish the `.d.ts`.

PopGoesTheWza: This could wait for a future PR IMHO. For an effective change, we need a new `ts2gas` release with updated dependency in `package.json`

grant:
Can't merge. Conflict.

PopGoesTheWza:
@grant I see the cause of the conflict and will push something to (tentatively) fix it

PopGoesTheWza:
@grant Once completed, I'll issue a new PR to fix the last linting error (which was in conflict)

> Consider using "forEach" instead of "map" as its return value is not being used here. (typescript:S2201)

grant:
Merged.

PR #431, opened 2018-11-30
Confirm overwrite using inquirer when manifest conflict occurs

takanakahiko:
Fixes #424 

This PR is according to https://github.com/google/clasp/issues/424#issuecomment-442899333

`clasp push` or `clasp push --watch`
-> If manifest conflict occurs, confirm overwrite using `inquirer`

`clasp push --force` or `clasp push --watch --force`
-> If manifest conflict occurs,  force pushing

- [x] `npm run test` succeeds.
- [x] `npm run lint` succeeds.
- [x] Appropriate changes to README are included in PR.

PR #618, opened 2019-05-16
(WIP) Glob match

### LINKED ISSUE #617: `.claspignore` does not handle `./` and `../` in paths
If the `rootDir` in the `.clasp.json` happens to contain `./` or `../` the .claspignore patterns fails to match anything.

This is in fact an issue within the `multimatch` module

## Expected Behavior

n.a.

## Actual Behavior

n.a.

1.
1.

## Specifications

- Node version (`node -v`): 10
- Version (`clasp -v`): latest
- OS (Mac/Linux/Windows): Mac

PopGoesTheWza:
Fixes #617 


PopGoesTheWza:
@grant please validate the principle before I go further

This PR allows to use rootDir like `../../build/gas/GasProject2`

I successfully tested `clasp push` with such projects.

It also corrects an issue when `filePushOrder` is empty and false-positive manifest changes

PopGoesTheWza:
@grant Please review and advise.

With this PR, there are potential breaking changes:

- `.claspignore` patterns now need to be relative to `rootDir`. Hence where rootDir is "dist", you no longer use `!dist/appsscript.json` but `!appsscript.json`
- It is not clear what `clasp status`' `untrackedFiles` should hold. Right now I made it return the `.claspignore` patterns followed by the list of ignored files. I am totally unsure why the first to pass siccesfuly and why the other two fails.

takanakahiko:
I think that it is better to refactor to other PR.
The more diffs there are, the higher the cost of the review.

ex : https://github.com/google/clasp/pull/618/commits/ea429a40c0c8f5d31b8936864cb6aa58127c351a

PopGoesTheWza:
@takanakahiko Sure. I will cancel this PR and issue distinct cleaner ones later.

Still I would like advise following my previous comment. :/

grant:
Yes. Relative to root dir.
Status should probably have tracked and ignored only.

Breaking change is fine.

src/files.ts
```diff
@@ -79,24 +81,32 @@ export async function getProjectFiles(rootDir: string = path.join('.', '/'), cal
...
-    ignoredFilePaths = ignoredFilePaths.concat(ignorePatterns);
 
     // Replace OS specific path separator to common '/' char for console output
     filePaths = filePaths.map((name) => name.replace(/\\/g, '/'));
+    filePaths.sort(); // Sort files alphanumerically
 
     // check ignore files
-    const ignoreMatches = multimatch(filePaths, ignorePatterns, { dot: true });
-    const intersection: string[] = filePaths.filter(file => !ignoreMatches.includes(file));
+    const acc = filePaths.reduce(
```
takanakahiko: You can create ignoreMatches and intersection without using `reduce` .

For example ...

```
const ignoreMatches = [];
const intersection = [];
filePaths.foreach( file => {
  const relative = path.relative(rootDir, file);
  if (multimatch(relative, ignorePatterns, { dot: true }).length === 0) {
    intersection.push(file);
  } else {
    ignoreMatches.push(file);
  }
});
```

src/commands/push.ts
```diff
@@ -95,5 +95,7 @@ const manifestHasChanges = async (): Promise<boolean> => {
   const remoteFiles = await fetchProject(scriptId, undefined, true);
   const remoteManifest = remoteFiles.find((file) => file.name === PROJECT_MANIFEST_BASENAME);
   if (!remoteManifest) throw Error('remote manifest no found');
-  return localManifest !== remoteManifest.source;
+  return localManifest !== remoteManifest.source
+    // if string comparison fail, double check after 'normalization'
+    && JSON.stringify(JSON.parse(localManifest)) !== JSON.stringify(JSON.parse(remoteManifest.source));
```
takanakahiko: Don't use JSON parse.

See https://github.com/google/clasp/pull/431#discussion_r237969173

src/files.ts
PopGoesTheWza: Sure. I just like `.reduce()` too much.

src/commands/push.ts
PopGoesTheWza: I had to add it as a failsafe. For some reason, I had a project where the string comparison was always failing. Adding the (silly) JSON stringify & parse allowed to workaround the issue.

I will remove it and re-test my projects.

src/files.ts
grant: Don't use `reduce`, it's not readable.
Use a `forEach` like @takanakahiko's example.

PopGoesTheWza:
Closing this PR for cleaner branches

PR #639, opened 2019-05-31
Pre pr cleanup

PopGoesTheWza:
Fixes no specific issues but improves types, fixes somes tests and upgrades packages


PopGoesTheWza:
@grant Here is a smaller PR with only the "small bits" not directly related with the config files. I will submit the proposed config files implementation in another PR (hoping I won't mess up)

src/index.ts
```diff
@@ -100,7 +100,7 @@ commander
   .description('Create a script')
   .option(
     '--type <type>',
-    'Creates a new add-on attached to a new Document, Spreadsheet, Presentation, or Form.',
+    'Creates a new Apps Script project attached to a new Document, Spreadsheet, Presentation, or Form.',
```
grant: Please revert. These are called 'add-ons' if they are attached to a script.

src/manifest.ts
```diff
@@ -22,7 +22,7 @@ export async function readManifest(): Promise<Manifest> {
   if (typeof rootDir === 'undefined') rootDir = DOT.PROJECT.DIR;
   const manifest = path.join(rootDir, PROJECT_MANIFEST_FILENAME);
   try {
-    return fs.readJsonSync(manifest, { encoding: 'utf-8' });
+    return fs.readJsonSync(manifest, { encoding: 'utf8' });
```
grant: Please use or create a constant for `'utf8'`.

src/index.ts
PopGoesTheWza: @grant I'm reverting.
Not that I want to argue, but doesn't Clasp addresses all the three types of scripts:

- standalone (including libraries?)
- bound to G Suite Documents (including add)ons)
- Web Apps

If so, reformulating this description would make sense, wouldn't it?

src/manifest.ts
PopGoesTheWza: @grant just to make sure: creating a `globals.ts` to hold `export const UTF8 = 'utf8'` is what you have in mind?

PopGoesTheWza:
@grant I'll let you set the resolved status if appropriate

src/index.ts
grant: Oh, I'd actually just change the comment to include all possibilities:

https://github.com/google/clasp/#examples-2

```
Creates a new Apps Script project attached to a new Document, Spreadsheet, Presentation, Form, or as a standalone script, web app, or API.
```

PopGoesTheWza: Your wish is my command.

src/manifest.ts
grant: Create a new constant in `files.ts`:

```ts
// @see https://nodejs.org/api/fs.html#fs_fs_readfilesync_path_options
export const FS_OPTIONS = { encoding: 'utf8' };
```

Use in the `src/` files:

https://github.com/google/clasp/search?q=utf&unscoped_q=utf

You don't have to use it in the test file.

PopGoesTheWza: Done

PopGoesTheWza:
@grant sorry that I introduced a circular dependency. Let me know how and when you want it to be fully addressed.

grant:
Thank you for the contribution.
Please have smaller PR in general.

PR #652, opened 2019-06-25
Logs detailed view

### LINKED ISSUE #638: Make clasp logs timestamps optional
## Expected Behavior
`clasp logs` should accept a `--timestamps` parameter to enable the periodic timestamping. Leaving it off should provide "clean" logs which are useful when logs are already timestamped.

## Actual Behavior
`clasp logs` always timestamps

mikearche:
Fixes #638 

I've added an optional parameter for the logs command called detailed. 

Timestamps are now disabled by default.
A user must use the `--detailed` command if they want timestamps now.

Works with the `--watch` command as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ clasp
 
 > **NOTE**: These commands require you to add your [Project ID](#projectid-optional).
 
-- [`clasp logs [--json] [--open] [--setup] [--watch] [--detailed]`](#logs)
+- [`clasp logs [--json] [--open] [--setup] [--watch] [--simplified]`](#logs)
 - [`clasp apis list`](#apis)
 - [`clasp apis enable <api>`](#apis)
 - [`clasp apis disable <api>`](#apis)
@@ -338,7 +338,7 @@ Prints out most recent the _StackDriver logs_. These are logs from `console.log`
 - `--open`: Open StackDriver logs in a browser.
 - `--setup`: Setup StackDriver logs.
 - `--watch`: Retrieves the newest logs every 5 seconds.
-- `--detailed`: Adds timestamps to the logs.
+- `--simplified`: Removes timestamps to the logs.
 
 #### Examples
 
@@ -351,7 +351,7 @@ ERROR      myFunction      error message
 - `clasp logs --json`
 - `clasp logs --open`
 - `clasp logs --watch`
-- `clasp logs --detailed`
+- `clasp logs --simplified`
 
 ### Run
 
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ clasp
 
 > **NOTE**: These commands require you to add your [Project ID](#projectid-optional).
 
-- [`clasp logs [--json] [--open] [--setup] [--watch]`](#logs)
+- [`clasp logs [--json] [--open] [--setup] [--watch] [--detailed]`](#logs)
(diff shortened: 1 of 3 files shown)


mikearche:
I signed it!

mikearche:
@grant thoughts?

grant:
Looks fine to me. Asked Steve to review.
Would need a minor version (x.1.x) bump.

sqrrrl:
This requires a major version bump as it is not backwards compatible. Anyone currently expecting timestamps will need to make a change when upgrading.

Why not invert the logic so the option is --no-timestamps? That would be backwards compatible with the current behavior.

mikearche:
Ah right, that makes much more sense. 
I'll make adjustments.

mikearche:
I used the terminology `--simplified` because it's similar terminology to the api documentation. 
[See Here](https://developers.google.com/apps-script/guides/logging)

README.md
```diff
@@ -338,18 +338,20 @@ Prints out most recent the _StackDriver logs_. These are logs from `console.log`
 - `--open`: Open StackDriver logs in a browser.
 - `--setup`: Setup StackDriver logs.
 - `--watch`: Retrieves the newest logs every 5 seconds.
+- `--simplified`: Removes timestamps to the logs.
```
grant: Isn't the new default without timestamps?

Shouldn't this be `--timestamps: Adds timestamps to the logs.`?

mikearche:
We are unable to make simplified the default view without making a major version bump. 
There may be users who are dependent upon the timestamps. 

We can definitely make it the default view if you would rather make a major version bump. I'll just revert my recent changes.

README.md
mikearche: Without a major version bump, this is not possible. I've updated the terminology though since "Removes timestamps to the logs" makes no sense.

README.md
```diff
@@ -338,18 +338,20 @@ Prints out most recent the _StackDriver logs_. These are logs from `console.log`
 - `--open`: Open StackDriver logs in a browser.
 - `--setup`: Setup StackDriver logs.
 - `--watch`: Retrieves the newest logs every 5 seconds.
+- `--simplified`: Removes timestamps from the logs.
 
 #### Examples
 
 ```text
 clasp logs
```
grant: We need to change this line if we are going to not show timestamps in the example.

```text
clasp logs --simplified
```

src/commands/logs.ts
```diff
@@ -16,10 +16,18 @@ const padEnd = require('string.prototype.padend');
  * @param cmd.open {boolean} If true, the command will open the StackDriver logs website.
  * @param cmd.setup {boolean} If true, the command will help you setup logs.
  * @param cmd.watch {boolean} If true, the command will watch for logs and print them. Exit with ^C.
+ * @param cmd.simplified {boolean} If true, the command will add timestamps to the logs.
```
grant: If **false**, will add timestamps

README.md
mikearche: I really didn't do my due diligence with the documentation 😥. Good catch.

src/commands/logs.ts
mikearche: Changed wordage to "If true, the command will remove timestamps from the logs".

grant:
Thanks for the contribution!

PR #662, opened 2019-07-08
Change how tests run so stack traces are accurate.

sqrrrl:
Use ts-node to run tests, change from *.js -> *.ts. Previously failed tests would produce inaccurate stack traces to the failed line in the tests.


grant:
Thanks!

Issue #676, opened 2019-07-25
Feature Request: Add command for checking login-status. 


diminishedprime:
## Expected Behavior

There is a cli command (`clasp login --status`) for finding out:

1) Is the user currently logged in
1) Who is the user logged in as

## Actual Behavior

No such command exists

## Details

I often work between a few different accounts, so I typically unconditionally run `clasp logout && clasp login` before working on code since I don't have a reliable way of testing which account I'm currently authenticated as. It would be nice if I could run a command to see if I needed to logout before doing it.

grant:
PRs welcome. May get to this eventually.

MattiasMartens:
PR: https://github.com/google/clasp/pull/685

HoldYourWaffle:
Should this be closed now that #685 has been merged?

MattiasMartens:
Yes it can be closed now!

PR #685, opened 2019-08-15
Login status

MattiasMartens:
Fulfils #676. `clasp login --status` which fetches user's email.

I wasn't able to get tests working locally in general. Most of them fail. I did set the environment variables but no luck. Would appreciate if someone else could confirm that tests pass.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -123,6 +123,9 @@ Did you provide the correct projectID?`,
 // Log messages (some logs take required params)
 export const LOG = {
   ASK_PROJECT_ID: `What is your GCP projectId?`,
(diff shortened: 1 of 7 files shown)


src/utils.ts
```diff
@@ -314,10 +317,21 @@ export function getAPIFileType(filePath: string): string {
 /**
  * Checks if the network is available. Gracefully exits if not.
  */
-export async function checkIfOnline() {
+export async function safeIsOnline() {
```
grant: Would rather combine `safeIsOnline` and `checkIfOnline`.
Have a param with `checkIfOnline` like `checkIfOnline(canThrowError = false)`.

Issue #441, opened 2018-12-07
npm install fails on Windows 10


libertybell95:
## Expected Behavior
Successful package install of clasp from NPM

## Actual Behavior
Upon execution of `npm i @google/clasp -g` in Windows CMD, install runs for a few seconds then outputs: 
```
npm ERR! Unexpected end of JSON input while parsing near '...DzCwGtdVc/u4hFqJRaI0d'

npm ERR! A complete log of this run can be found in:
npm ERR!       AppData\Roaming\npm-cache\_logs\2018-12-07T18_06_25_632Z-debug.log
```

Debug log: [2018-12-07T18_06_25_632Z-debug.log](https://github.com/google/clasp/files/2658395/2018-12-07T18_06_25_632Z-debug.log)

Have attempted to run install with and without administrator permissions without success. Also attempted `npm i -g grpc @google/clasp --unsafe-perm` and received identical error message.

## Steps to Reproduce the Problem

  1. Execute `npm i @google/clasp -g` in Command Line

## Specifications

  - Node version (`node -v`): 10.14.1
  - Version (`clasp -v`): N/A
  - OS (Mac/Linux/Windows): Windows 10

grant:
Perhaps try different versions:

```
npm i -g @google/clasp@1.6.0
# or
npm i -g @google/clasp@1.5.0
```

Maybe upgrade Node to 11.4.0?
Clear the npm cache like @kenchan0130 suggested?

libertybell95:
**Resolved**

Steps Taken:
1. Uninstalled Node.JS and reinstalled with 11.4.0.
2. Performed regular 'npm i @google/clasp -g' with no success
3. Cleared and verified npm cache successfully
4. Ran npm install for clasp v1.6.0 successfully

Issue #190, opened 2018-05-26
npm install is inconsistent


grant:
I'm not able to consistently install clasp. I sometimes get these errors:

```
sudo npm i -g @google/clasp
/usr/local/bin/clasp -> /usr/local/lib/node_modules/@google/clasp/index.js

> grpc@1.12.2 install /usr/local/lib/node_modules/@google/clasp/node_modules/grpc
> node-pre-gyp install --fallback-to-build --library=static_library

node-pre-gyp WARN Pre-built binaries not installable for grpc@1.12.2 and node@8.11.2 (node-v57 ABI, unknown) (falling back to source compile with node-gyp)
node-pre-gyp WARN Hit error EACCES: permission denied, mkdir '/usr/local/lib/node_modules/@google/clasp/node_modules/grpc/src/node'
gyp WARN EACCES user "root" does not have permission to access the dev dir "/var/root/.node-gyp/8.11.2"
gyp WARN EACCES attempting to reinstall using temporary dev dir "/usr/local/lib/node_modules/@google/clasp/node_modules/grpc/.node-gyp"
gyp WARN EACCES user "root" does not have permission to access the dev dir "/usr/local/lib/node_modules/@google/clasp/node_modules/grpc/.node-gyp/8.11.2"
...
```

# Configs

```
npm -v
5.6.0
node -v
v8.11.2
```

# Command

```
sudo npm i -g @google/clasp
/usr/local/bin/clasp -> /usr/local/lib/node_modules/@google/clasp/index.js
+ @google/clasp@1.3.1
updated 1 package in 11.211s
```

```
sudo npm uninstall -g @google/clasp
```

# Possible Solutions

This seems to not give the errors:
`sudo npm i -g @google/clasp --allow-root`

Another option would be to remove `grpc`, but I believe it is needed with a different npm module.

`sudo npm i -g @google/clasp --unsafe-perm`
seems to always work too.

However, running clasp gives this error still:

```
/usr/local/lib/node_modules/@google/clasp/node_modules/grpc/src/grpc_extension.js:45
    throw error;
    ^

Error: The gRPC binary module was not installed. This may be fixed by running "npm rebuild"
Original error: Cannot find module '/usr/local/lib/node_modules/@google/clasp/node_modules/grpc/src/node/extension_binary/node-v57-darwin-x64-unknown/grpc_node.node'
[...]

PR #702, opened 2019-09-06
700 local clasp

### LINKED ISSUE #700: Clasp does not need to be installed globally
currently only global install of clasp is documented. This implies that upgrading clasp can break existing projects.

## Expected Behavior

One can install clasp as a local dependency and use nom scripts (or npx’ on the command line) to use the local version of clasp

npm install @google/clasp
npx clasp —version

1.
1.

## Specifications

- Node version (`node -v`):
- Version (`clasp -v`):
- OS (Mac/Linux/Windows):

PopGoesTheWza:
Fixes #700


README.md
```diff
@@ -63,7 +63,7 @@ First download `clasp`:
 
 ```sh
 npm install -g @google/clasp
-# Alternatively: sudo npm i -g grpc @google/clasp --unsafe-perm
+# Alternatively: sudo npm install -g grpc @google/clasp --unsafe-perm
```
PopGoesTheWza: @grant any context as for when/why one try to `sudo` and add a 3rd party `grpc` package?

PopGoesTheWza:
@grant There maybe some spelling errors :/

PopGoesTheWza:
Approval status?

README.md
```diff
@@ -517,14 +517,53 @@ Specifies the file extension for **local** script files in your Apps Script proj
 
 Specifies the files that should be pushed first, useful for scripts that rely on order of execution. All other files are pushed after this list of files.
 
+## Using clasp as a local dependency
```
grant: I don't support this on the main README.

Create a separate `doc/` page if we want to add this guide.

README.md
grant: > @grant any context as for when/why one try to `sudo` and add a 3rd party `grpc` package?

I don't know. It may be an old workaround. We need someone to completely uninstall Node, all libraries, and install from scratch and verify it works.

README.md
```diff
@@ -517,14 +517,53 @@ Specifies the file extension for **local** script files in your Apps Script proj
...
+From the command line, within your project, use the `npx` tool in order to use the **local** version of `clasp` rather than any global one.
+
+```sh
+npx clasp --version
+```
+
 ## Troubleshooting
 
-### Node Version
+### NodeJS Version
```
grant: The main README needs less text, not more.

I oppose this change.

README.md
PopGoesTheWza: Will do

README.md
PopGoesTheWza: Just to be sure I get it: Node is préfères over NodeJS right? If so, I will revert.

grant: No, this whole section should be reverted. There are no useful changes.

PopGoesTheWza:
Ok boss

README.md
PopGoesTheWza: I have clasp running on MacOS and Windows on "clean" node instances without need of this `grpc`.
If it solved any issue with clasp, it is likely a side-effect it seems.

README.md
PopGoesTheWza: I beg to differ ;) supported version of node is now up to date. And the upgrade instruction no longer "pollutes" your global node_modules with a "one-shot" `n` module.
I would personally drop the whole paragraph, on the reasoning that checking and upgrading node version is already covered length on SO and the rest of the internet.

README.md
PopGoesTheWza: I found original mention of `grpc` in #190 
From what I can gather, some of the dependencies at the time used grpc but could require you to manually update to latest version.

Currently grpc is no longer mentionned in `node_modules` nor `package-lock.json`

I suggest we wipe this outdated "fix" (cf. #441 )

grant:
Sorry, I'm not the maintainer of this project anymore.
Deferring to @sqrrrl.

docs/README.md
```diff
@@ -4,5 +4,6 @@ This folder contains advanced guides for `clasp`.
 
 - [Develop clasp](/docs/develop.md): Build, Run, and Test the CLI.
 - [Run](/docs/run.md): Execute Apps Script remotely with `clasp run`.
+- [Running loaclly](/docs/running-locally.md): Install and use clasp as a local dependency.
```
grant: There is a typo here

PopGoesTheWza: Thanks. I usines and merged a PR fixing it.

PR #739, opened 2020-02-22
V8 engine support

### LINKED ISSUE #735: build option for V8 runtime
https://developers.google.com/apps-script/guides/v8-runtime/migration

Do you have any plan to support less-transformed codes for v8 runtime?

PopGoesTheWza:
Fixes #735 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ language: node_js
 node_js:
 - "12"
 - "10"
-- "8"
 env:
   global:
   - SCRIPT_ID=1EwE84eZCSBPcaAiJzCnDjmxMVnLQrDyhSKq1oZY6q-3x4BIDHgQefCnL
--- a/README.md
+++ b/README.md
@@ -61,6 +61,12 @@ You can also try clasp in Gitpod, a one-click online IDE for GitHub:
 - Easily add custom Google OAuth scopes
 - [And more...](docs/run.md)
 
+**- V8 support** take advantage of the performance boost of Chrome JavaScript engine:
+
+- Every ES2019 features (except ES modules)
+- Edit your `appsscript.json` manifest to choose between the **Rhino** and **V8** engines
+- Typescript users should update their `tsconfig.json` with the `"target": "2019"` compiler option
+
 ## Install
 
 First download `clasp`:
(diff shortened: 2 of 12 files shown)


PopGoesTheWza:
@shinout : please review & test this PR

if OK, we'll ask grant for a release

shinout:
@PopGoesTheWza 
Also, I have to say that I failed to push ESNext codes to GAS even using this branch of clasp. As I repeatedly said, it's because of GAS server which parses pushed codes as non-ESNext ones. I think it's too early to release this version until GAS accepts ESNext codes via HTTP. Have you ever encountered a similar situation?

PopGoesTheWza:
> Otherwise, tsconfig files cannot be found in monorepo-structured projects.

@shinout do you have a sample monorepo to test?

I personally use this [template](https://github.com/PopGoesTheWza/ts-gas-project-starter) for single repo/multiple GAS targets. I plan on updating it to support V8 engine.

PopGoesTheWza:
> until GAS accepts ESNext codes via HTTP

@shinout I am unsure what you mean. I had no time for testing clasp push of ES2019 yet but intend to do so this weekend.

shinout:
@PopGoesTheWza 
Here is a sample monorepo directory structure.
This is the most typical one when using [lerna](https://github.com/lerna/lerna).
```
(monorepo root)
├── tsconfig.ts
├── packages
│   └── gas-package01
│       └── index.ts
```

The essence is that the main directory of each package is two or more directories. The current implementation only considers the parent directory of a given project. `ts. findConfigFile()` looks up ancestors recursively.
https://github.com/microsoft/TypeScript/blob/4a34294908bed6701dcba2456ca7ac5eafe0ddff/src/compiler/program.ts#L4

 > I am unsure what you mean

Sorry about that. Let me explain more precisely. The command `clasp push` uses `googleapis`'s method `projects.updateContent()`. It accesses to [this](https://developers.google.com/apps-script/api/reference/rest/v1/projects/updateContent) REST API, which seems to validate pushed codes before updating contents. The validation fails when we push ESNext codes, then the update cannot be done. I suppose the validation logic at their REST API server still doesn't accept ESNext grammar. Thus, even if this version of clasp is released, we won't be able to push ESNext codes.

PopGoesTheWza:
@grant ready for release.

I drafted the release notes <https://github.com/google/clasp/releases/edit/untagged-477cea24e1e150fb3fc4>

PR #755, opened 2020-03-21
Fix the default command's output

kalmi:
Since the commander upgrade in c805769c04b6dcd8c8ea3eb0e20856655e2dc255, clasp was printing the following when encountering an unknown command:
`Unknown command "clasp [object Object]"`

### DIFF
--- a/src/commands/defaultCmd.ts
+++ b/src/commands/defaultCmd.ts
@@ -1,9 +1,10 @@
 import { ERROR, logError } from '../utils';
+import { Command } from 'commander';
 
 /**
  * Displays a default message when an unknown command is typed.
  * @param command {string} The command that was typed.
  */
-export default async (command: string): Promise<void> => {
+export default async (_: Command, command: string): Promise<void> => {
   logError(null, ERROR.COMMAND_DNE(command));
 };
--- a/tests/commands/defaultCmd.ts
+++ b/tests/commands/defaultCmd.ts
@@ -1,0 +1,15 @@
+import { spawnSync } from 'child_process';
+import { expect } from 'chai';
+import { describe, it } from 'mocha';
+import { CLASP } from '../constants';
+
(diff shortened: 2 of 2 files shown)


PopGoesTheWza:
FYI, @kalmi I'll review this PR tomorrow

PopGoesTheWza:
@kalmi wouldn't the [custom event handler](https://github.com/tj/commander.js#custom-event-listeners) approach be more appropriate?

kalmi:
It probably would be. The `*       Any other command is not supported` line is not very useful in the output of `clasp help`.

I wanted to avoid changing the index.ts file to avoid a merge conflict with your long-running WIP branch.

PopGoesTheWza:
@kalmi don't worry about merge conflict at this point.

Being confined with two youngsters (4 & 6 yo) I can barely handle my regular work load, so contributing to this Google project is difficult and a luxury.

Current code base would benefit from some major refactoring. Some primary point of concern:
- async & promises: implementation is potentially flawed in several places. Commander's `.parseAsync()` should be used instead of `.parse()`, making `handleError()` useless in the process
- error handling is minimal and lacking in some places
- some (too many IMHO) clasp commands underwent extensive kludgy fixes/hacks and have become unreadable/unintelligible 
- likewise helpers like `utils`, `files`, `manifest` have grown in some monstuous mess with circular dependencies
- unit testing is clumsy, incomplete, failing on Windows

Maybe I should open issues detailing the above points and ask for help.

kalmi:
Good luck with youngsters. I hope everyone is understanding.

I think that some of these would make "good first issues", and that label might encourage more contribution.  There must be quite a few childless developers stuck at home right now looking for something to do. 🥺.. or maybe not, not sure how ppl work in this regard in the current climate.

I won't be able to contribute code for next few days, but I hope to be back soon.

PR #820, opened 2020-12-19
Updating documentation solving #819

### LINKED ISSUE #819: Support for developing using ES Modules
## Expected Behavior
Documentation describing a setup where one can use ES Modules (import/export-statements)

## Actual Behavior
Currently not in scope

N/A

## Specifications

PR is coming...

atti187:
Fixes #819

- [n/a] `npm run test` succeeds.

### DIFF
--- a/docs/esmodules.md
+++ b/docs/esmodules.md
@@ -78,4 +78,4 @@ node_modules/**
 
 Now you can push using your normal `clasp push`-command!
 
-This should be enough to start developing a project using ES Modules.
+This should be enough to start developing a project using ES Modules. A complete example repo can be cloned from [atti187/esmodules](https://github.com/atti187/esmodules).
--- a/docs/esmodules.md
+++ b/docs/esmodules.md
@@ -60,7 +60,7 @@ module.exports = {
 };
 ```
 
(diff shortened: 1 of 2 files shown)


contributorpw:
I think it would be great to specify the entry point of the call to `clasp` or some example workflow when `clasp` is called.

atti187:
@contributorpw - to my understanding, clasp picks every file not being .ignored, so it's business as usual. Nevertheless, I made some clarifications.

PopGoesTheWza:
A detailed practical example would be nice. Maybe something like:
- "we need advanced date manipulation with moment.js"
- retrieve ES module and install it (npm add ?)
- minimal index.ts using moment.js 

Also a paragraph to clarify any limitation about ES Modules which can be "imported"; also the difference with npm (typically commonjs) modules

atti187:
I think that would rather be part of an example repo that could be linked to. I can see if I can set it up.

As for third party dependencies, I'm not sure if and how that'd work with the run-time restrictions of App Scripts - depending on what they're using. Also handling all plugins needed to cope with different module patterns etc is probably out of scope for this PR.

Essentially what this handles is a way to structure your own code.

atti187:
Using moment actually worked with the small test I did! Added a reference to my example repo...

atti187:
Any updates on this? Has anyone tried it out?

PopGoesTheWza (approved):

PR #828, opened 2021-03-18
fix: exit the process after the command action is performed. Closes #805

### LINKED ISSUE #805: Avoiding Ctrl+c exits to every call
## Expected Behavior
Running any call (e.g. `clasp push`) should push to the server, then exit, returning to the normal terminal prompt.

## Actual Behavior
Running any call (e.g. `clasp push`)  pushes correctly to the server, but the terminal remains unresolved, or unclosed properly. It is responsive, but will not return to the normal terminal prompt without forcefully exiting with Ctrl+c. 

1. `clasp push`
1. Hangs until Ctrl-c input

## Specifications

- Node version (`node -v`): 14.9.0
- Version (`clasp -v`): 2.3.0
- OS (Mac/Linux/Windows): Linux Ubuntu 18.04.4

nilp0inter:
This fixes #805.

I had to disable [this lint rule](https://eslint.org/docs/rules/no-process-exit) which I think does not apply in this particular case. But maybe I am wrong, I am not a JS developer.

- [X] `npm run test` succeeds. (With some errors not introduced by this change)
- [X] `npm run lint` succeeds. (With some errors not introduced by this change)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -367,4 +367,6 @@ if (args.length === 0) {
       console.error('Unknown error', error);
     }
   }
+  /* eslint-disable no-process-exit*/
+  process.exit();
 })();


nilp0inter:
@googlebot I signed it!

PopGoesTheWza:
@nilp0inter I do not favor forcing `process.exit()` because:

- it is considered bad practice
- the root cause for delayed exits has not yet been identified

nilp0inter:
That's ok. 😊

Issue #832, opened 2021-04-19
clasp clone ignores rootDir option if there is already a .clasp.json in a parent directory


VAggrippino:
> (Note: Non-breaking issues are likely not to be prioritized. Please consider a PR in addition to your issue)

## Expected Behavior
`clasp clone aaaBBBzzz --rootDir .` clones a project and writes the `.clasp.json` in the specified directory.

## Actual Behavior
`clasp clone aaaBBBzzz --rootDir .` clones a project in the current directory, but updates the `.clasp.json` file in a higher level directory rather than writing one in the specified directory.

## Steps to Reproduce the Problem

1.`clasp clone <project-id> --rootDir .` in a subdirectory of another clasp project.

## Specifications

- Node version (`node -v`): 10.19.0
- Version (`clasp -v`): 2.3.0
- OS (Mac/Linux/Windows): Linux

sqrrrl:
Fixed in 2.4.1

PR #865, opened 2021-08-09
Refactor config paths, fix CLI options

### LINKED ISSUE #859: Project arguments are ignored in command line
## Expected Behavior

Using the command line arguments `-P` or `--project` should allow you to specify a clasp project file according to the help text: `path to a project file or to a folder with a '.clasp.json' file`.

## Actual Behavior

These arguments are ignored.

1. Move your `.clasp.json` file to a new name like `myClasp.json`
2. Run `clasp -P myClasp.json versions`
3. You will see an error that `.clasp.json` was not found because the `-P` arg did not properly use the provided file

## Specifications

- Version (`clasp -v`): 2.4.0

sqrrrl:
fix: Honor --project and other CLI options for overriding config file paths

Fixes #859 & #832.  Removes PathProxy as no longer needed -- simpler to inline the path resolution in Config and not have the extra layer of indirection.


src/commands/logout.ts
```diff
@@ -1,40 +1,16 @@
 import {Conf} from '../conf.js';
-import {DOTFILE} from '../dotfile.js';
-import {hasOauthClientSettings} from '../utils.js';
+import fs from 'fs';
```
PopGoesTheWza, line 2: shouldn't it be fs-extra?

src/index.ts
```diff
@@ -52,13 +52,14 @@ import versions from './commands/versions.js';
 import {Conf} from './conf.js';
 import {PROJECT_NAME} from './constants.js';
 import {spinner, stopSpinner} from './utils.js';
+import fs from 'fs';
```
PopGoesTheWza, line 55: fs-extra?

PopGoesTheWza (changes requested):
@sqrrrl LGTM

I suggest the use of 'fs-extra' over 'fs'.

src/commands/logout.ts
sqrrrl: done

src/index.ts
sqrrrl: done

PopGoesTheWza (approved):
LGTM

FYI, I just published updated `dotf` and `ts2gas` packages.

PopGoesTheWza (approved):

PR #892, opened 2021-11-30
Add delete command

### LINKED ISSUE #799: Add command to delete Apps Script projects
For testing + staging purposes we'd like to be able to spin up ephemeral Apps Script projects which exist only to test a PR. We can currently create the projects with `clasp create` but there is no matching `clasp delete` to clean up afterwards.

## Expected Behavior
There is a `clasp delete` command to delete the Apps Script project with a particular ID (or read from .clasp.json).

## Actual Behavior
N/A

N/A

## Specifications
N/A

fantonangeli:
Closes #799 


sqrrrl:
Closing for two reasons:

* Large refactoring in 3.x means this needs to be reworked. Just took a stab at it and it's straightforward, but testing leads to the second main issue....
* Clasp does not currently request the necessary oauth scopes to implement this correctly. Would work for scripts created by clasp, but would fail with insufficient permission for any other script and the process to request the full drive scope needs to be done well ahead of this.

fantonangeli:
Hi @sqrrrl I was working on the refactoring of this PR.
Regarding you point N. 2: Do you mean it's not possible to use this scope?
`https://www.googleapis.com/auth/drive`

sqrrrl:
It's possible, but to enable for clasp's built-in oauth client requires pre-approval. It's a restricted scope too, so subject to some conditions and may not get approved. That said, clasp already uses some restricted scopes so may not be a big deal, but if we want to include this that work will have to be done ahead of time. Adding the drive scope to clasp before that's done would cause errors during login.

sqrrrl:
WIll look into it, though to set expectations this can take anywhere on the order of weeks or even a few months, so no rush on the code :) But it's relatively easy to re-implement, not at all worried about that part.

Issue #934, opened 2022-08-24
Error: Looks like you offline


jvanalst:
> (Note: Non-breaking issues are likely not to be prioritized. Please consider a PR in addition to your issue)

## Expected Behavior

That it actually work without error. Tried it with both a Personal and Enterprise Google Accounts. Both had the same issue. 

## Actual Behavior

> % clasp login
> Logging in globally…
> 🔑 Authorize clasp by visiting this url:
> https://accounts.google.com/o/oauth2/v2/auth?access_type=offline&scope=...
> Authorization successful.
> 
> Default credentials saved to: /Users/vanalsti/.clasprc.json.
> 
> % clasp clone 1vSgFkRCFRM52k4VL_dQROAUWzmR6DqLRkUkZpD7jvvFYewoXhPuGKzjT
> ⠙ Cloning files…script.google.com ✖
> console.developers.google.com ✖
> console.cloud.google.com ✖
> drive.google.com ✖
> Error: Looks like you are offline.
> 
> % clasp create foobar
> script.google.com ✖
> console.developers.google.com ✖
> console.cloud.google.com ✖
> drive.google.com ✖
> Error: Looks like you are offline.
>
> % clasp login --status
> script.google.com ✖
> console.developers.google.com ✖
> console.cloud.google.com ✖
> drive.google.com ✖
> Error: Looks like you are offline.
>
> % clasp login
> Warning: You seem to already be logged in *globally*. You have a ~/.clasprc.json
> Logging in globally…
> 🔑 Authorize clasp by visiting this url:
> https://accounts.google.com/o/oauth2/v2/auth?access_type=offline&scope=...
> 
> Authorization successful.
> 
> % clasp login --status
> console.developers.google.com ✖
> script.google.com ✖
> console.cloud.google.com ✖
> drive.google.com ✖
> Error: Looks like you are offline.

## Steps to Reproduce the Problem

1. clasp login
1. clasp create/clone/etc.

## Specifications

- Node version (`node -v`): v16.14.1
- Version (`clasp -v`): 2.4.1
- OS (Mac/Linux/Windows): Mac 12.5

jvanalst:
All doesn't work work with Clasp 2.4.1 on Node 14.

jvanalst:
Possibly Related: https://github.com/google/clasp/issues/872

jvanalst:
Downgraded clasp to 2.3.1 seems to work.

Issue #930, opened 2022-07-16
error push


eka05:
> (Note: Non-breaking issues are likely not to be prioritized. Please consider a PR in addition to your issue)

## Expected Behavior
can not push

## Steps to Reproduce the Problem
Help me please

PS E:\Clasp Google Script> clasp push

PS E:\Clasp Google Script> clasp push
drive.google.com ✖
Error: Looks like you are offline.

No valid E:\Clasp Google Script.clasp.json project file. You may need to create or clone a project first.
PS E:\Clasp Google Script> clasp push -w
console.cloud.google.com ✖
drive.google.com ✖
Error: Looks like you are offline.

## Specifications

- Node version (`node -v`): v16.16.0
- Version (`clasp -v`): 2.4.1
- OS (Mac/Linux/Windows): windows 10

eko-develops:
@eka05  Try logging in with clasp first if you have not yet.
https://github.com/google/clasp#login

After logging in with clasp, you will need to create/clone the project first before being able to push. 
https://github.com/google/clasp#create
https://github.com/google/clasp#clone

Issue #913, opened 2022-02-26
clasp push -w randomly crashes


andymelt:
## Expected Behavior
I would expect `clasp push -w` to sit indefinitely without crashing and dumping crash information.

## Actual Behavior
After random amounts of time —sometimes it happens within minutes after just a couple of pushes and sometimes it will run for hours without issue— clasp will crash and dump information about the cause. Here is an example of the output after the crash:

```
Error: The `onCancel` handler was attached after the promise settled.
    at onCancel (/opt/homebrew/lib/node_modules/@google/clasp/node_modules/p-cancelable/index.js:48:12)
    at makeRequest (/opt/homebrew/lib/node_modules/@google/clasp/node_modules/got/dist/source/as-promise/index.js:38:13)
    at Request.<anonymous> (/opt/homebrew/lib/node_modules/@google/clasp/node_modules/got/dist/source/as-promise/index.js:143:17)
    at Object.onceWrapper (node:events:646:26)
    at Request.emit (node:events:526:28)
    at Timeout.retry (/opt/homebrew/lib/node_modules/@google/clasp/node_modules/got/dist/source/core/index.js:1270:30)
    at Timeout.retry (/opt/homebrew/lib/node_modules/@google/clasp/node_modules/wtfnode/index.js:197:27)
    at listOnTimeout (node:internal/timers:559:17)
    at processTimers (node:internal/timers:502:7)
node:events:504
      throw er; // Unhandled 'error' event
      ^

RequestError: Timeout awaiting 'request' for 25000ms
    at ClientRequest.<anonymous> (/opt/homebrew/lib/node_modules/@google/clasp/node_modules/got/dist/source/core/index.js:962:65)
    at Object.onceWrapper (node:events:646:26)
    at ClientRequest.emit (node:events:538:35)
    at ClientRequest.origin.emit (/opt/homebrew/lib/node_modules/@google/clasp/node_modules/@szmarczak/http-timer/dist/source/index.js:43:20)
    at TLSSocket.socketErrorListener (node:_http_client:442:9)
    at TLSSocket.emit (node:events:526:28)
    at emitErrorNT (node:internal/streams/destroy:164:8)
    at emitErrorCloseNT (node:internal/streams/destroy:129:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
    at Timeout.timeoutHandler (/opt/homebrew/lib/node_modules/@google/clasp/node_modules/got/dist/source/core/utils/timed-out.js:36:25)
[...]

Issue #909, opened 2022-02-15
Online check breaks clasp commands due to bug in got


kenkku:
Got, which is used in clasp through `is-reachable`, has an unresolved bug that causes an error in some circumstances. sindresorhus/got#1489

## Expected Behavior
Commands succeed with exit code 0 when there is no errors

## Actual Behavior
```
?master ~/github/[removed]> npx clasp status
Not ignored files:
└─ src/appsscript.json
└─ src/Code.ts

Ignored files:

Error: The `onCancel` handler was attached after the promise settled.
    at onCancel (/Users/kenkku/github/[removed]/node_modules/p-cancelable/index.js:48:12)
    at makeRequest (/Users/kenkku/github/[removed]/node_modules/got/dist/source/as-promise/index.js:38:13)
    at Request.<anonymous> (/Users/kenkku/github/[removed]/node_modules/got/dist/source/as-promise/index.js:143:17)
    at Object.onceWrapper (events.js:520:26)
    at Request.emit (events.js:400:28)
    at Timeout.retry (/Users/kenkku/github/[removed]/node_modules/got/dist/source/core/index.js:1270:30)
    at Timeout.retry (/Users/kenkku/github/[removed]/node_modules/wtfnode/index.js:197:27)
    at listOnTimeout (internal/timers.js:557:17)
    at processTimers (internal/timers.js:500:7)
exit 1
```

## Steps to Reproduce the Problem
I don't know what the steps to reproduce are, as the underlying bug seems to be a race condition. I have reproduced it on two different Macs, but it did not happen on a colleague's Mac

## Specifications

- Node version (`node -v`): v14.19.0 (also tried with 16)
- Version (`clasp -v`): 2.4.1
- OS (Mac/Linux/Windows): Mac

PR #936, opened 2022-09-01
fix: remove online check

grant:
Removes online check. I believe this was added from a user's PR from long ago, but it has caused many user errors over the years. The CLI shouldn't even special case to check if the client is online, the client should just directly try. The [implementation](https://github.com/google/clasp/blob/aa4197c00193e83b71f4b696e67730bbb48fdaf9/src/utils.ts#L186-L226) is kind of scary when you look at it (not attempted to change / remove).

Fixes https://github.com/google/clasp/issues/872
Fixes https://github.com/google/clasp/issues/934
Fixes https://github.com/google/clasp/issues/886
Fixes https://github.com/google/clasp/issues/913
Fixes https://github.com/google/clasp/issues/909
Fixes https://github.com/google/clasp/issues/930

R=@sqrrrl

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -12,7 +12,7 @@ import type {ReadonlyDeep} from 'type-fest';
 import {ClaspError} from './clasp-error.js';
 import {DOTFILE} from './dotfile.js';
 import {ERROR, LOG} from './messages.js';
-import {checkIfOnlineOrDie, getOAuthSettings} from './utils.js';
+import {getOAuthSettings} from './utils.js';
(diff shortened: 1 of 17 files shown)


sqrrrl (approved):

kenkku:
This error prevented us from using clasp. Thanks @grant!

grant:
> This error prevented us from using clasp. Thanks @grant!

Glad you found this tool and this PR useful @kenkku ! Looks like a bunch of folks got unblocked reading the comments.

Thanks to @sqrrrl for the quick merge.

PR #1044, opened 2025-03-24
Fix: Normalize file paths in file operations for Windows compatibility

hd140283:
Fixes # (no related issue)

- [ ] Appropriate changes to README are included in PR. (N/A)

## Summary
This PR fixes file path normalization in file operations to ensure compatibility with Windows environments. 

## Details
- Corrected path normalization logic to handle Windows-specific file paths.
- Updated test cases to verify correct path normalization on Windows.

## Testing
All tests and lint checks were executed successfully on a Windows environment.

### DIFF
--- a/src/core/files.ts
+++ b/src/core/files.ts
@@ -197,7 +197,7 @@ export class Files {
         const localPath = path.relative(process.cwd(), path.join(contentDir, filename));
         const resolvedPath = path.relative(contentDir, localPath);
         const parsedPath = path.parse(resolvedPath);
-        let remotePath = path.format({dir: normalizePath(parsedPath.dir), name: parsedPath.name});
+        let remotePath = normalizePath(path.format({dir: parsedPath.dir, name: parsedPath.name}));
 
         const type = getFileType(localPath, fileExtensionMap);
         if (!type) {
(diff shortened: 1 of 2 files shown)


src/core/files.ts
```diff
@@ -197,7 +197,7 @@ export class Files {
         const localPath = path.relative(process.cwd(), path.join(contentDir, filename));
         const resolvedPath = path.relative(contentDir, localPath);
         const parsedPath = path.parse(resolvedPath);
-        let remotePath = path.format({dir: normalizePath(parsedPath.dir), name: parsedPath.name});
+        let remotePath = normalizePath(path.format({dir: parsedPath.dir, name: parsedPath.name}));
```
Nu11u5, line 200: Should the remote filepath format be dependent on the local OS?

hd140283: Just to clarify:

The remote path is normalized using [`normalize-path`](https://www.npmjs.com/package/normalize-path), so it's always converted to POSIX format (`/`), regardless of the local OS. This ensures consistent behavior across different environments.

That logic was implemented in my PR and has already been merged into the `master` branch — but it hasn’t been released to NPM yet.

If you're using the latest published version of `@google/clasp`, the change may not be included yet.

hd140283: I just moved `normalizePath()` to wrap the result of `path.format()`, since `path.format()` is OS-dependent.

Issue #950, opened 2022-11-24
Feat : clasp login with service account


Identity-labs:
> (Note: Non-breaking issues are likely not to be prioritized. Please consider a PR in addition to your issue)

## Expected Behavior
Login with clasp with service account key

## Actual Behavior

Logging in locally…

Authorizing with the following scopes:
https://www.googleapis.com/auth/userinfo.email
https://www.googleapis.com/auth/script.external_request
https://www.googleapis.com/auth/script.webapp.deploy

NOTE: The full list of scopes your project may need can be found at script.google.com under:
File > Project Properties > Scopes

Error retrieving access token: TypeError: Cannot destructure property 'client_id' of 'options.creds.installed' as it is undefined.

## Steps to Reproduce the Problem

1. Create service account into GCP
2. Generate keys & download it
3. npx @google/clasp login --creds generated-key.json

## Specifications

- Node version (`node -v`): v14.21.1
- Version (`clasp -v`): 2.3.1 or 2.4.2
- OS (Mac/Linux/Windows): Mac or Linux

sqrrrl:
Looks sort of possible, but with some caveats.

Apps script API works with service accounts *if* the script is owned by a user and shared with the service account. However, a service account can not *own* the script itself nor enable use of the Apps Script API, so some commands like create & run won't work using service account credentials.  Still need to check if version & deploy work.

But if the goal is to lock down CI, it may be viable as only a handful of commands need to work. And since scripts would have to be explicitly shared with the service account, it would help limit access.

sqrrrl:
Available in 3.0-alpha, please give it a try and see if it meets needs. 

Note that service accounts can't own scripts, but can read/write if shared. But at least now should be possible to use service accounts in a CI to update scripts

tlkiong:
I tried the above but I got `Cannot read properties of undefined (reading 'redirect_uris')`

legoguy1000:
> I ran into the following error after running:
> 
> ```
> npx @google/clasp login --creds generated-key.json
> ```
> 
> error:
> 
> > Cannot read properties of undefined (reading 'redirect_uris')

I"m also getting this error and would like to use a service account to use in CICD.

jenul-ferdinand:
Any updates on this, still not working for me.

jddcef:
[...]

PR #1100, opened 2025-11-28
Update auth.ts

BrodBlox09:
Throw error in the case the user-provided credentials file does not include keys "installed" or "web" to prevent unclear "Cannot read properties of undefined (reading 'redirect_uris')" error. https://github.com/google/clasp/issues/950#issuecomment-2715462870 no longer seems to be accurate, since the code does not seem to allow for service account JSON key files.

Amends #950 partially (makes error more clear)

### DIFF
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -226,7 +226,7 @@ async function saveOauthClientCredentials(store: CredentialStore, userKey: strin
 }
 
 /**
- * Creates an aunthorized oauth2 client with the given credentials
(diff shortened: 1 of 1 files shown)


BrodBlox09:
I signed the CLA using my Google account, but I'm not sure it will work since I made the commit with `99990717+BrodBlox09​@users.noreply.github.com`.

BrodBlox09:
Attempted to update author email to the one I signed the CLA with

PR #1139, opened 2026-03-30
fix(auth): remove hardcoded OAuth client secret from published npm package (CWE-798)

g0w6y:
## Summary

This PR fixes the sensitive data exposure (CWE-798) where `@google/clasp` shipped its OAuth2 client secret in plaintext inside every published npm tarball (`build/src/auth/oauth_client.js`).

**Changes**
- Removed `DEFAULT_CLASP_OAUTH_CLIENT_SECRET` completely from the codebase
- Updated `auth.ts` and `file_credential_store.ts` to no longer reference the secret
- Default login now runs as a **public client** (client ID only)
- Explicitly added `build/` to `.npmignore`
- Preserved full backward compatibility for `--creds` and existing V1 `.clasprc.json` files (legacy credentials now correctly use public-client mode)

**Why this matters**
- Eliminates the hardcoded credential vulnerability
- Makes the default authentication flow secure by default
- No breaking changes for any user

**Verification**

```
npm pack
tar -xf google-clasp-*.tgz
grep -r "v6V3fKV_zWU7iw1DrpO1rknX" package/   # should return nothing
```

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,4 @@
 src
 test
 .*
+build
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -25,7 +25,7 @@ import {AuthorizationCodeFlow} from './auth_code_flow.js';
 import {CredentialStore} from './credential_store.js';
 import {FileCredentialStore} from './file_credential_store.js';
 import {LocalServerAuthorizationCodeFlow} from './localhost_auth_code_flow.js';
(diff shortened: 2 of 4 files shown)


g0w6y:
@sqrrrl Hi please review

sqrrrl:
Thanks. Just added PKCE support in another PR to round this out.

PR #1140, opened 2026-03-30
Fix Path Traversal Vulnerability in pull command via .clasp.json srcDir

l3tchupkt:
Previously, `srcDir` was resolved using `path.resolve()` without enforcing that the resulting path stays within the project root directory. This allowed paths containing `../` to escape the project directory.

As a result, when running `clasp pull`, files fetched from the Apps Script project could be written outside the project directory.

### Impact

In scenarios where a user clones or interacts with an untrusted repository containing a crafted `.clasp.json`, this could lead to arbitrary file writes outside the project directory.

Depending on the target path and environment, this may impact user files or configuration.

### Root Cause

`path.resolve(projectRoot, srcDir)` was used without validating that the resolved path remains inside the project root. Existing safeguards only validated file paths relative to `contentDir`, not the `contentDir` itself. 

### Fix

* Reused the existing `isInside` utility from `src/core/files.ts`
* Added a validation check in `initClaspInstance` to ensure the resolved `contentDir` remains within the project root
* Throws an error if traversal outside the project root is detected

### Example (before fix)

```json
{
  "srcDir": "../../.ssh"
}
```

Running `clasp pull` would resolve this outside the project directory and write files there.

### Result

This change ensures that all file operations performed by `clasp pull` remain constrained within the project directory, preventing path traversal via `srcDir`.

### Reported by

Lakshmikanthan K (letchupkt)

---

### DIFF
--- a/src/core/clasp.ts
+++ b/src/core/clasp.ts
@@ -25,7 +25,7 @@ import JSON5 from 'json5';
 import splitLines from 'split-lines';
 import stripBom from 'strip-bom';
 import {getUserInfo} from '../auth/auth.js';
-import {Files} from './files.js';
+import {Files, isInside} from './files.js';
 import {Functions} from './functions.js';
 import {Logs} from './logs.js';
 import {Project} from './project.js';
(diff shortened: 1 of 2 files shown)


sqrrrl:
This is a breaking change. There's no need to prevent people from *deliberately* pointing to sources outside the location of the clasp config file. I don't think it's very common, but also not at all clear that there is any vulnerability as a result.

The previous fix for path traversal prevents a malicious script from breakout out of the content root. That should still hold true here where the content root is the absolute path of what srcdir resolves to.

Can you explain what particular concern this is aiming to fix? 

Again, I don't expect this at all to be a common configuration, but unless ithere's an explicit security issues it'll have to wait until the 4.x release.

sqrrrl (approved):

l3tchupkt:
@sqrrrl Thanks for the review and merge, appreciate it.

I’ll also work on a short security write-up and PoC demonstrating the real-world impact (arbitrary file write via untrusted .clasp.json) so it’s clearly documented for users.

Let me know if you’d prefer it as a security advisory draft.

PR #1147, opened 2026-04-09
fix(security) : Restrict credential file path to home directory

l3tchupkt:
fixes(#1148)
Prevent credential theft by validating --auth paths and blocking 
symlink attacks that could exfiltrate OAuth tokens.

Changes:
- Enforce credential files must be within home directory
- Detect and block symlinks pointing outside home (~/safe -> /tmp)
- Use O_NOFOLLOW | O_EXCL for atomic credential writes
- Maintain 0o600 permissions on credential files

Fixes: Credential path injection and symlink bypass vulnerabilities

- [ x] `npm run test` succeeds.
- [x ] `npm run lint` succeeds.


l3tchupkt:
@sqrrrl please review!!

sqrrrl:
Have to think about this one.  May be too restrictive given the nature of the tool.

There are legitimate use cases for secrets outside the home directory. Some CI/CD platforms allow mounting secrets as files. Currently have the --auth flag documented as deprecated, but it still may have some utility.

Can you split this into two parts? I think most of the changes are OK (don't follow symlinks, maintaining correct file permissions, etc.) but the path part I'm less confident about.

Alternatively, may be worth adding an option flag like --allow-unsafe-paths for this and other recent changes that may break some less common use cases.

l3tchupkt:
@sqrrrl Makes sense 👍

Goal was to prevent credential leaks since unsafe paths or symlinks can expose tokens (like in `/tmp` or CI mounts). But yeah, blocking might break real use cases.

I’ll split it:

* keep safe writes + no symlinks + strict perms
* change path check to warning
* add `--allow-unsafe-paths` if needed

Let me know your thoughts.

sqrrrl:
sounds good

sqrrrl:
Resolved in a separate change, closing as obsolete

PR #1159, opened 2026-07-01
fix(push): error on non-existent filePushOrder entries instead of failing silently

### LINKED ISSUE #984: A silent failure occurs if a file path listed in `filePushOrder` does not exist
> (Note: Non-breaking issues are likely not to be prioritized. Please consider a PR in addition to your issue)

## Expected Behavior

If a file path listed in `filePushOrder` does not exist, I would expect `clasp push` to fail with a clear error message.

## Actual Behavior

It just skips the missing file and continues as if nothing went wrong.

1. Add a path to a non-existing file to `filePushOrder`
1. Run `clasp push`
  1. Observe that the output to stdout and stderr includes no error message about the non-existing file path
1. Run `echo $?`
   1. Observe that the exit code of `clasp push` is `0` but should have been non-zero.

## Specifications

- Node version (`node -v`): `v20.8.1`
- Version (`clasp -v`): `2.4.2`
- OS (Mac/Linux/Windows): macOS 13.6 22G120

gunjanjaswal:
Fixes #984.

`filePushOrder` in the manifest lets you order pushed files. If it lists a path that doesn't exist locally, `clasp push` currently succeeds silently with exit code `0` — the missing entry is just skipped.

`Files.checkMissingFilesFromPushOrder` already computed the missing entries but never returned or used them (its own JSDoc noted it "doesn't do anything with the result"), and the method was never called anywhere. This wires it in:

- the method now returns the missing `filePushOrder` entries;
- `push()` calls it before uploading and throws if any entry is missing, so the command fails with a clear message and a non-zero exit — the behavior the issue asks for.

Existing behavior is unchanged; the check only fires when a `filePushOrder` path has no matching local file.

## Test

Added a case in `test/core/files.ts`: with `filePushOrder: ['config.js', 'missing.js']` and no `missing.js` on disk, `push()` now rejects. The throw happens before any API call, so no network mock is needed.

Happy to switch to a `console.warn` instead of throwing if you'd prefer a softer failure, but the issue expects a non-zero exit.


gunjanjaswal:
@googlebot I signed it!

sqrrrl:
OK, but technically a breaking change, will hold for a little bit until ready to work on 4.x

gunjanjaswal:
Makes sense, no rush from my side. Happy to rebase it whenever the 4.x work kicks off, or fold it into a broader 4.x breaking-changes batch if that's easier to manage. Thanks for taking a look.

